PR #3, opened 2018-02-05
[wpmlcore-5172] Making some methods from WPML_VC_Website_Builder_Tran…

adrianosferreira:
…slatable_Nodes reusable

### DIFF
--- a/src/st/compatibility/class-wpml-page-builders-translatable-nodes.php
+++ b/src/st/compatibility/class-wpml-page-builders-translatable-nodes.php
@@ -1,0 +1,135 @@
+<?php
+
+abstract class WPML_Page_Builders_Translatable_Nodes {
+
+	/**
+	 * @var array
+	 */
+	protected $nodes_to_translate;
+
+	/**
+	 * @param string|int $node_id
+	 * @param array $settings
+	 *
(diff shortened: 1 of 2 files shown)


src/st/compatibility/class-wpml-page-builders-translatable-nodes.php
```diff
@@ -0,0 +1,135 @@
+<?php
+
+abstract class WPML_Page_Builders_Translatable_Nodes {
```
brucepearson: @adrianosferreira Should this have `implements IWPML_Page_Builders_Translatable_Nodes`?

adrianosferreira: @brucepearson I don't think so. Child classes like `WPML_VC_Website_Builder_Translatable_Nodes` will implement `IWPML_Page_Builders_Translatable_Nodes` so they know which methods must be declared.

brucepearson: @adrianosferreira Are you sure because this looks like it class does implement the interface.

adrianosferreira: @brucepearson done.

PR #4, opened 2018-02-15
[wpmlcore-5204] Moved code for saving translated post body from Gitla…

adrianosferreira:
…b to Github

### DIFF
--- a/src/st/class-page-builder-integration.php
+++ b/src/st/class-page-builder-integration.php
@@ -5,6 +5,8 @@
  */
 class WPML_Page_Builders_Integration {
 
+	const STRINGS_TRANSLATED_PRIORITY = 10;
+
 	/** @var WPML_Page_Builders_Register_Strings */
 	private $register_strings;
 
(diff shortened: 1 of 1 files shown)


adrianosferreira:
@OnTheGoSystems/wpml please review this. When moving things from ST to Github somehow this was not moved.

src/st/class-page-builder-integration.php
```diff
@@ -5,6 +5,8 @@
  */
 class WPML_Page_Builders_Integration {
 
+	const UPDATE_TRANSLATED_POST_PRIORITY = 10;
```
sciamannikoo: @adrianosferreira shouldn't the constant reflect the name of the action isntead of the callback?

sciamannikoo:
Except for the constant's name, there isn't much to see here, which is good :)

src/st/class-page-builder-integration.php
adrianosferreira: @sciamannikoo makes sense, fixed.

PR #5, opened 2018-02-21
[wpmltm-1847] Added attribute in tag XML config file for encoding con…

adrianosferreira:
…dition

A new attribute `encoding-condition` has been added in shortcode tag for the XML config file. This attribute will allow us to conditionally encode a tag defined to be encoded.

For now it only works in this format `option:my_option_name=my_option_value`. Which means that if such option has such value, then the tag will be encoded.


adrianosferreira:
@BruceOnTheGo please check it.

adrianosferreira:
@brucepearson please check my latest commit https://github.com/OnTheGoSystems/wpml-page-builders/pull/5/commits/72b45344d7c97bd8daed751509f3ab6f66ba72f5

I had to add `html_entity_decode` because the base64 decoded string had entities.

dimitrismitsis:
@OnTheGoSystems/wpml one more review please

Issue #6, opened 2018-03-01
preg_replace() can give-up on long strings and this results in empty translated page


GeorgeBotsev:
If we have a very long string that we need to process, then preg_replace can fail and return NULL.
This results in the fact that the contents of the translation are blank in the resultant translated page/post.
Here is the error log:
`[01-Mar-2018 10:44:50 UTC] PHP Warning:  preg_replace(): Compilation failed: regular expression is too large at offset 69790 in /var/www/cats.com/wp-content/plugins/sitepress-multilingual-cms/vendor/wpml/page-builders/src/st/strategy/shortcode/class-wpml-pb-update-shortcodes-in-content.php on line 91
[01-Mar-2018 10:44:50 UTC] PHP Stack trace:
[01-Mar-2018 10:44:50 UTC] PHP   1. {main}() /var/www/cats.com/wp-admin/admin-ajax.php:0
[01-Mar-2018 10:44:50 UTC] PHP   2. do_action() /var/www/cats.com/wp-admin/admin-ajax.php:97
[01-Mar-2018 10:44:50 UTC] PHP   3. WP_Hook->do_action() /var/www/cats.com/wp-includes/plugin.php:453
[01-Mar-2018 10:44:50 UTC] PHP   4. WP_Hook->apply_filters() /var/www/cats.com/wp-includes/class-wp-hook.php:310
[01-Mar-2018 10:44:50 UTC] PHP   5. call_user_func_array:{/var/www/cats.com/wp-includes/class-wp-hook.php:286}() /var/www/cats.com/wp-includes/class-wp-hook.php:286
[01-Mar-2018 10:44:50 UTC] PHP   6. wpml_save_job_ajax() /var/www/cats.com/wp-includes/class-wp-hook.php:286
[01-Mar-2018 10:44:50 UTC] PHP   7. WPML_TM_Editor_Job_Save->save() /var/www/cats.com/wp-content/plugins/wpml-translation-management/inc/ajax.php:46
[01-Mar-2018 10:44:50 UTC] PHP   8. WPML_TM_Editor_Save_Ajax_Action->run() /var/www/cats.com/wp-content/plugins/wpml-translation-management/classes/menu/translation-editor/class-wpml-tm-editor-job-save.php:12
[...]

PR #7, opened 2018-03-12
Use a predefined subroutine in order to decrease the portion which do…

adrianosferreira:
…es the matching

This fix is needed because we are exceeding memory limit when doing preg_replace for really big strings.

The idea is to create a sub regex routine which will hold the expression together with the long string. Moreover, the portion which in fact does te matching will be smaller, e.g: `(?&text)`.

https://github.com/OnTheGoSystems/wpml-page-builders/issues/6


andreasciamanna:
@adrianosferreira shall we close this PR?

adrianosferreira:
@andreasciamanna yes.

PR #8, opened 2018-03-19
[wpmlcore-5257] Adjusted text-domain to 'sitepress'

andreasneuber:
Reason: code has moved from WPML String Translation to WPML Core plugin


adrianosferreira:
@andreasneuber looks good. Will you do the same in Beaver Builder / Elementor?

Issue #10, opened 2018-04-18
Performance hit when saving the translation editor with a lot of page builder fields.


strategio:
WPML Page Builders tries to update the translated post on each string that is translated. This is quite strange because there's already a stack where we store the packages to be refreshed in the `shutdown` action, see the code in `\WPML_PB_String_Translation::new_translation` which calls `\WPML_PB_String_Translation::save_translations_to_post`.

This `\WPML_PB_String_Translation::save_translations_to_post` method is called by `\WPML_PB_Integration::do_shutdown_action`.

I think we should just get rid of this, or I am missing something...

Reference: wpmlcore-5304

strategio:
This was fixed with https://github.com/OnTheGoSystems/wpml-page-builders/pull/11, we can close it now.

PR #11, opened 2018-04-19
Fix performance hit when saving the translation editor with a lot of

strategio:
page builder fields

We were updating the translated post content on each package string
update. This was already done in the shutdown for all attached packages
that were modified.

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-5304
https://github.com/OnTheGoSystems/wpml-page-builders/issues/10

### DIFF
--- a/src/st/class-wpml-pb-string-translation.php
+++ b/src/st/class-wpml-pb-string-translation.php
@@ -27,9 +27,6 @@ class WPML_PB_String_Translation {
 			$package = $this->factory->get_wpml_package( $package_id );
 			if ( $package->post_id && $this->strategy->get_package_kind() === $package->kind ) {
 				$this->add_package_to_update_list( $package, $language );
-				if ( defined( 'DOING_AJAX' ) && DOING_AJAX || defined( 'XMLRPC_REQUEST' ) && XMLRPC_REQUEST ) {
-					$this->save_translations_to_post();
-				}
 			}
 		}
 	}


src/st/class-wpml-pb-string-translation.php
```diff
@@ -27,9 +27,6 @@ public function new_translation( $translated_string_id ) {
 			$package = $this->factory->get_wpml_package( $package_id );
 			if ( $package->post_id && $this->strategy->get_package_kind() === $package->kind ) {
 				$this->add_package_to_update_list( $package, $language );
-				if ( defined( 'DOING_AJAX' ) && DOING_AJAX || defined( 'XMLRPC_REQUEST' ) && XMLRPC_REQUEST ) {
-					$this->save_translations_to_post();
-				}
```
andreasciamanna: @strategio are you sure this condition wasn't added there for a specific reason?

Because that's what it looks like to me. It seems like we had to solve a problem (maybe the compatibility with a different page builder) so we added these lines.

adrianosferreira: Better to ask @BruceOnTheGo @brucepearson about this.

strategio: @andreasciamanna, @adrianosferreira, quoting myself in the YT ticket:
>Looking at the Git history in ST, I can see that this action to update the post has been moved to the shutdown in wpmlst-963. At this time, we should have removed this call in `\WPML_PB_String_Translation::new_translation` I guess.
>
>Then later, we also started to update the post when handling an XML-RPC request. Looking at the original tickets (wpmlst-1210 / wpmltm-1759), I think wpmlst-1210 was not needed (and tested this successfully this morning).

strategio:
@BruceOnTheGo, can you give it a look please?

PR #12, opened 2018-05-11
Replace only attribute value of the current attribute

adrianosferreira:
The old regex pattern was not looking at the attribute name when replacing the its value. It was causing problems with attributes that have same value however different attribute name. I've added the attribute name in the regex for replacing only what we want.

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-5288


adrianosferreira:
@brucepearson @brucepearson please review this.

adrianosferreira:
@BruceOnTheGo @brucepearson it was not so easy but I could write it, please check once more.

PR #14, opened 2018-07-03
[wpmltm-2670] Add a hook for translations into the post

adrianosferreira:
Recently we've stopped updating the post on each string package iteration (when updating) because it was causing performance issues, so we choose to only update it once in the “shutdown” hook. However, this change is avoiding third parties to properly use the hook `wpml_pro_translation_completed`, because when they hook into it, the translated post body does not have any content in the post body yet (because shutdown didn’t get triggered at that time).

Now, I’ve added a new action to run as soon as all string packages are updated and, of course, before the shutdown action. This new action is actually added inside of the Page Builders callback of `wpml_pro_translation_completed`.

### DIFF
--- a/src/st/class-wpml-pb-integration.php
+++ b/src/st/class-wpml-pb-integration.php
@@ -100,6 +100,7 @@ class WPML_PB_Integration {
 		add_action( 'save_post', array( $this, 'queue_save_post_actions' ), PHP_INT_MAX, 2 );
 		add_action( 'icl_st_add_string_translation', array( $this, 'new_translation' ), 10, 1 );
 		add_action( 'shutdown', array( $this, 'do_shutdown_action' ) );
+		add_action( 'wpml_pb_finished_adding_string_translations', array( $this, 'save_translations_to_post' ) );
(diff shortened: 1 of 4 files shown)


strategio:
@adrianosferreira, does it still makes sense to update the post in the `shutdown` if it was already updated on `wpml_pb_finished_adding_string_translations`? Maybe we can remove the `shutdown` action if we already updated. Saying this because we have a heavy process on post update and I think it's worth avoiding a double update (this can easily save us 1 or 2 seconds sometimes).

adrianosferreira:
@strategio no, there are other important things done in the `shutdown` action, not only `save_translations_to_post`:

```
$this->save_translations_to_post();

		remove_action( 'icl_st_add_string_translation', array( $this, 'new_translation' ), 10, 1 );

		foreach( $this->save_post_queue as $post_id => $post ) {
			$this->register_all_strings_for_translation( $post );
		}
```

I prefer to let the save_translation_to_post there as we probably have other actions doing more things between the `wpml_pro_translation_completed` and `shutdown` actions.

PR #16, opened 2018-08-06
Implement class for detecting if a particular page is a page builder …

adrianosferreira:
…page

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlpb-148


brucepearson:
@adrianosferreira This looks ok but I would like to see how/where it is called so I can see if it makes sense in context.

adrianosferreira:
@BruceOnTheGo ok, it is on the way!

PR #19, opened 2018-08-10
Allow replacing long strings in page builder shortcodes

strategio:
To fix the issue https://github.com/OnTheGoSystems/wpml-page-builders/issues/6.

`preg_replace` throws an error when the number of characters is too large.

We can consider that we have a very little risk of collision for long strings so we can use a simple `str_replace` in that case.

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-5470

### DIFF
--- a/src/st/strategy/shortcode/class-wpml-pb-update-shortcodes-in-content.php
+++ b/src/st/strategy/shortcode/class-wpml-pb-update-shortcodes-in-content.php
@@ -2,6 +2,8 @@
 
 class WPML_PB_Update_Shortcodes_In_Content {
 
+	const LONG_STRING_THRESHOLD = 5000;
+
 	/** @var  WPML_PB_Shortcode_Strategy $strategy */
 	private $strategy;
 	/** @var WPML_PB_Shortcode_Encoding $encoding */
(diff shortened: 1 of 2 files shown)


src/st/strategy/shortcode/class-wpml-pb-update-shortcodes-in-content.php
```diff
@@ -80,17 +80,25 @@ private function update_shortcode_attributes( $shortcode_data ) {
...
 	private function replace_string_with_translation( $block, $original, $translation, $is_attribute = false, $attr = '' ) {
 		$translation = apply_filters( 'wpml_pb_before_replace_string_with_translation', $translation, $is_attribute );
+		$new_block   = $block;
 
-		$new_block = $block;
 		if ( $translation ) {
-			if ( $is_attribute && $attr ) {
-				$pattern   = '/' . $attr . '=(["\'])' . preg_quote( $original, '/' ) . '(["\'])/';
-				$new_block = preg_replace( $pattern, $attr . '=${1}' . $translation . '${2}', $block );
+			$is_text_too_long_for_regex = mb_strlen( $original ) > 5000;
```
andreasciamanna: @strategio I would move this to a private method and the `5000` to a constant.

andreasciamanna:
@BruceOnTheGo I think you may want to give a look at this.

src/st/strategy/shortcode/class-wpml-pb-update-shortcodes-in-content.php
strategio: Done @andreasciamanna

strategio:
@andreasciamanna, the reason we use `preg_replace` instead of `str_replace` is to a better control. This prevents unexpected replacement (e.g. a string which has the same value as an attribute name) or multiple replacements (like we fixed in ST MR 617 / wpmlst-1536). This should not be a problem to use `str_replace` for very long strings.

PR #20, opened 2018-08-17
Add support for `raw-html` attribute

strategio:
- This new attribute is now parsed in the config file.
- If raw html is set to "not translate" in the settings, the shortcodes
with `raw-html` set to 1 will not be added.

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlpb-154


strategio:
Please review @OnTheGoSystems/wpml

PR #22, opened 2018-08-29
Add shared logic and interface for media translation

strategio:
- Reused part of `WPML_Page_Builders_Update_Translation` moved to `WPML_Page_Builders_Update` which I will now use in composition.
- Introduced the interface `IWPML_PB_Media_Update`.
- Introduced `WPML_Page_Builders_Update_Media` which can be used as a base class for page builders using API.
- Introduced the filter `wmpl_pb_get_media_updaters` which will provide the `IWPML_PB_Media_Update` instances to run the media translation. This filter might also be used later in WPML Media.
- Added the media translate logic during shutdown action in PB Integration.

I started using this code in https://github.com/OnTheGoSystems/wpml-page-builders-beaver-builder/pull/8.

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlpb-160

### DIFF
--- a/src/media/class-wpml-page-builders-update-media.php
+++ b/src/media/class-wpml-page-builders-update-media.php
@@ -1,0 +1,43 @@
+<?php
+
+abstract class WPML_Page_Builders_Update_Media implements IWPML_PB_Media_Update {
+
+	/** @var WPML_Page_Builders_Update $pb_update */
+	private $pb_update;
+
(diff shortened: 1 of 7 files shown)


strategio:
Closing @brucepearson. I created a new PR with part 1 and 2, and I also created an abstraction here to load the media uploader hooks.

See https://github.com/OnTheGoSystems/wpml-page-builders/pull/24.

PR #24, opened 2018-09-03
Base code to load and run page builders media updaters

strategio:
- Reused part of `WPML_Page_Builders_Update_Translation` moved to `WPML_Page_Builders_Update` which I will now use in composition.
- Introduced the interface `IWPML_PB_Media_Update`,  `IWPML_PB_Media_Update_Factory`, and `IWPML_PB_Media_Node_Iterator`.
- Introduced `WPML_Page_Builders_Update_Media` and `WPML_Page_Builders_Media_Hooks` which can be used in composition.
- Introduced the filter `wmpl_pb_get_media_updaters` which will provide the `IWPML_PB_Media_Update` instances to run the media translation. This filter might also be used later in WPML Media.
- Added the media translate logic during shutdown action in PB Integration.

See how this code is used in https://github.com/OnTheGoSystems/wpml-page-builders-beaver-builder/pull/10.

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlpb-160

### DIFF
--- a/src/media/class-wpml-page-builders-media-hooks.php
+++ b/src/media/class-wpml-page-builders-media-hooks.php
@@ -1,0 +1,35 @@
+<?php
+
+class WPML_Page_Builders_Media_Hooks implements IWPML_Action {
+
+	/** @var IWPML_PB_Media_Update_Factory $media_update_factory */
+	private $media_update_factory;
+
+	/** @var string $page_builder_slug */
+	private $page_builder_slug;
+
+	/**
+	 * WPML_Page_Builders_Media_Hooks constructor.
+	 *
+	 * @param IWPML_PB_Media_Update_Factory $media_update_factory
+	 * @param string                        $page_builder_slug
+	 */
+	public function __construct( IWPML_PB_Media_Update_Factory $media_update_factory, $page_builder_slug ) {
+		$this->media_update_factory = $media_update_factory;
+		$this->page_builder_slug    = $page_builder_slug;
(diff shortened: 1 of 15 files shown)


strategio:
@brucepearson, please review :)

src/media/class-wpml-page-builders-media-translate-helper.php
```diff
@@ -0,0 +1,85 @@
+<?php
+
+class WPML_Page_Builders_Media_Translate_Helper {
```
brucepearson: @strategio I suggest you drop the `_Helper` from the name. A class with the name `Helper` suggest to me that it's a class that doesn't have a clear role.

src/media/interface-iwpml-pb-media-node-iterator.php
```diff
@@ -0,0 +1,6 @@
+<?php
+
+interface IWPML_PB_Media_Node_Iterator {
```
brucepearson: @strategio I don't think `_Iterator` makes sense here. The doesn't not iterate over anything that I can see. Isn't it just a node?

strategio: @brucepearson, actually it should be `IWPML_PB_Media_Nodes_Iterator` (with a "s" for Nodes) because traverse all the media nodes and try to translate it.

adrianosferreira:
@strategio +1.

PR #26, opened 2018-09-14
WIP - Improvements in tests and CI

andreasciamanna:
This MR started as a solution to a bunch of report issues coming from Scrutinizer and ended up as a way to mainly get rid of `otgs/unit-tests-framework` dependency (a private repository which blocks contributors from running tests and adds a lot of unneeded additional dependencies).

The library has been replaced by `otgs/phpunit-tools` (public library) which provides the minimum features (tools) to run PHPUnit tests in this kind of projects.

By doing that, I'm now able to:
- Run tests independently any private repository
- Run phpcs for checking standards as well as compatibility issues
- Get cleaner (hopefully) Scrutinizer reports

### DIFF
--- a/.scrutinizer.yml
+++ b/.scrutinizer.yml
@@ -1,12 +1,12 @@
 ---
 checks:
-  php:
-    code_rating: true
-    duplication: true
+    php:
+        code_rating: true
+        duplication: true
 filter:
     excluded_paths:
-        - "tests/"
-        - "vendor/"
+    - "tests/"
+    - "vendor/"
 coding_style:
     php:
         indentation:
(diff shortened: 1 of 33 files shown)


adrianosferreira:
> This MR started as a solution to a bunch of report issues coming from Scrutinizer and ended up as a way to mainly get rid of `otgs/unit-tests-framework` dependency (a private repository which blocks contributors from running tests and adds a lot of unneeded additional dependencies).

@andreasciamanna I don't think you should remove it, It is a public repository: https://packagist.org/packages/otgs/unit-tests-framework

> 
> The library has been replaced by `otgs/phpunit-tools` (public library) which provides the minimum features (tools) to run PHPUnit tests in this kind of projects.
> 
> By doing that, I'm now able to:
> 
> * Run tests independently any private repository
> * Run phpcs for checking standards as well as compatibility issues
> * Get cleaner (hopefully) Scrutinizer reports

andreasciamanna:
@adrianosferreira I didn't remember we made it public, but still think it's a library we don't need in this project.
If you look at the changes, you'll see that by just changing the base class with a simpler one (and only because I didn't bother creating the stub class in the few places where it was used), and adjusting a few tests with what I think is also a better code, tests run just fine.

For instance, we were using `mock_all_core_functions` everywhere either for no reason or for laziness.

Anyway, I'm still testing these changes. I'll ask for a review as soon as I'm confident with it :)

andreasciamanna:
@OnTheGoSystems/wpml this PR is completely optional, but as I've tried to explain, I think it would help in many ways:

- Fewer or smaller Composer dependencies
- Scrutinizer code inspection with less false positives
- Cleaner tests

If we decide to proceed, it would be better to continue merging on Pierre's branch, just because I didn't realize what I was doing before it was too late :)

Mind that this PR changes nothing in the code base anyway: it's all about tests and even for them, what changes is only the way we mock or stub, not the logic of the test itself, except for one case where I added more coverage.

andreasciamanna:
I forgot to mention that you can compare the difference of inspection issues:
- With my changes: https://github.com/OnTheGoSystems/wpml-page-builders/pull/26/checks
- Without: https://github.com/OnTheGoSystems/wpml-page-builders/pull/25/checks

strategio:
I think it's a good move @andreasciamanna. +1

andreasciamanna:
@strategio as the target of this change is https://github.com/OnTheGoSystems/wpml-page-builders/pull/25 (which you own), I'll let you deal with the merging.

This PR can be merged as is (no squashing required).

PR #31, opened 2018-09-21
Changes for Cornerstone integration

adrianosferreira:
- Implemented a way to check for a function when trying to check if a page builder is active. This was needed for Cornerstone, because the plugin itself apparently has no constants. I didn't want to check for a X Theme constant because the integration is for Cornerstone plugin.
- Defined another abstract method in data settings class for optionally copy the post body.

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-5811


adrianosferreira:
@OnTheGoSystems/wpml please review

andreasciamanna:
I think the PR is fine if we stick purely to the scope of the PR.

But I was put down by the inherited quality of the code in `WPML_Page_Builders_Defined` which, if we keep dismissing, won't get better on its own.

adrianosferreira:
> I think the PR is fine if we stick purely to the scope of the PR.
> 
> But I was put down by the inherited quality of the code in `WPML_Page_Builders_Defined` which, if we keep dismissing, won't get better on its own.

@andreasciamanna I think your MR makes sense. I will cherry-pick it here and let other developers review it.

adrianosferreira:
@andreasciamanna I've cherry-picked your commit and tested, it seems good. However, I will describe better in test steps what to assert in the manual tests. Also, I've fixed tests for your change.

src/st/compatibility/class-wpml-page-builders-update-translation.php
```diff
@@ -56,6 +56,5 @@ protected function get_translation( WPML_PB_String $string ) {
 		return $string;
 	}
 
-	abstract protected function update_strings_in_modules( array &$data_array );
```
andreasciamanna: @adrianosferreira you've removed this method which is called in this very class as line #37.

andreasciamanna:
@adrianosferreira I've made additional changes. Please look at my last three commits in this PR.

Other than that, it looks good to me.

andreasciamanna:
@adrianosferreira other than the removed method (see my other comment), the PR seems fine.

But that missing method should be restored, I think. I wonder why we don't get errors in our tests. Probably, we are missing some coverage.

src/st/compatibility/class-wpml-page-builders-update-translation.php
adrianosferreira: @andreasciamanna this is an abstract class, so each lib has its own update translation class. This class is covered on each lib, so if it would fail it wouldn't be here, but in the lib (e.g Cornerstone). Look that it is well implemented there and that's why it didn't fail:

https://github.com/OnTheGoSystems/wpml-page-builders-cornerstone/blob/wpmlcore-5811/src/class-wpml-cornerstone-update-translation.php

Of course, I shouldn't remove this, it was done by mistake. This method should be in the abstract class simply to work like an interface, so child classes know it should implement that. Thanks for pointing it out.

adrianosferreira:
@andreasciamanna I've pushed changes, please check.

src/st/compatibility/interface-iwpml-page-builders-data-settings.php
```diff
@@ -44,5 +44,13 @@ public function prepare_data_for_saving( array $data );
...
+	 * @return void
+	 */
 	public function add_hooks();
-}
\ No newline at end of file
+
+	/**
+	 * @return bool
+	 */
+	public function should_copy_post_body();
```
brucepearson: @adrianosferreira The problem with adding a new function is that all the classes that implement this now need to be updated.

I suggest that the call to `get_fields_to_copy` can also return `body`. Then the code can check for this and copy the body accordingly.

adrianosferreira: @brucepearson I already created PR for Elementor and Beaver Builder. Do you think I should change it?

adrianosferreira: @brucepearson what you said makes sense. I will change that.

adrianosferreira:
@brucepearson I've done like you suggested. Please check.

src/st/compatibility/class-wpml-page-builders-update.php
```diff
@@ -47,6 +47,16 @@ private function save_data( $post_id, $fields, $data ) {
 	 */
 	private function copy_meta_fields( $translated_post_id, $original_post_id, $meta_fields ) {
```
brucepearson: @adrianosferreira This function should now be renamed to `copy_fields` and `$meta_fields` renamed to `$fields`

src/st/compatibility/class-wpml-page-builders-update.php
```diff
@@ -47,6 +47,16 @@ private function save_data( $post_id, $fields, $data ) {
...
 		foreach ( $meta_fields as $meta_key ) {
+			if ( 'post_content' === $meta_key ) {
+				$original_post = get_post( $original_post_id );
+				wp_update_post( array(
+					'ID' => $translated_post_id,
+					'post_content' => $original_post->post_content,
+				) );
+
+				continue;
```
brucepearson: @adrianosferreira I would remove the `continue` and put the copy of the meta values in the `else`

adrianosferreira: @brucepearson done.

src/st/compatibility/class-wpml-page-builders-update.php
brucepearson: @adrianosferreira Did you do this change?

adrianosferreira: @brucepearson no, I missed that one, sorry. Fixing in a moment.

andreasciamanna:
There are conflicts here I'm not sure how to fix. @adrianosferreira that's something you should handle.

PR #33, opened 2018-09-24
 Improve shortcodes definition structure in wpml-config.xml

strategio:
The improve structure is as below:

```xml
<!-- Existing shortcodes definition (for backward compatibility)-->
<shortcode>
    <!-- Translate string in the content -->
    <!-- Only in strings settings -->
    <tag>et_pb_foo</tag>
</shortcode>
<shortcode>
    <!-- Translate link attribute as a string, translate string in the content -->
    <!-- Only in strings settings -->
    <tag>et_pb_bar</tag>
    <attributes>
        <attribute type="link">link</media-attribute>
    </attributes>
</shortcode>

<!-- New shortcodes definition -->
<shortcode>
    <!-- Convert media IDs in `gallery_ids` attribute, and ignore the shortcode content -->
    <!-- Only in media settings -->
    <tag ignore-content="1">et_pb_gallery</tag>
    <attributes>
        <attribute type="media-ids">gallery_ids</media-attribute>
    <attributes>
</shortcode>
<shortcode>
    <!-- Convert the media URL in the content, translate some string attributes, convert media IDs in `src` attribute -->
    <!-- Both in strings and media settings -->
    <tag type="media-url">et_pb_image</tag>
    <attributes>
        <attribute>title_text</attribute>
        <attribute>alt</attribute>
        <attribute type="link">link</attribute>
        <attribute type="media-ids">src</attribute>
    </attributes>
</shortcode>
```

I added as second commit to simplify the code in the strings settings. Indeed we were having 2 separate logics for single and multiple attributes which was making the code hard to follow and debug.

This goes with some changes in the XSD file: https://git.onthegosystems.com/wpml/sitepress-multilingual-cms/merge_requests/2081.

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-5867

### DIFF
--- a/src/media/shortcodes/class-wpml-page-builders-media-shortcodes.php
+++ b/src/media/shortcodes/class-wpml-page-builders-media-shortcodes.php
@@ -2,7 +2,6 @@
 
 class WPML_Page_Builders_Media_Shortcodes {
 
-	const ALL_TAGS = '\w+';
 	const TYPE_URL = 'media-url';
 	const TYPE_IDS = 'media-ids';
 
@@ -31,7 +30,7 @@ class WPML_Page_Builders_Media_Shortcodes {
 	public function translate( $content )  {
 		foreach ( $this->config as $shortcode ) {
 			$shortcode = $this->sanitize_shortcode( $shortcode );
-			$tag_name  = $this->get_tag_name( $shortcode );
+			$tag_name  = isset( $shortcode['tag']['name'] ) ? $shortcode['tag']['name'] : '';
 
 			if ( ! empty( $shortcode['attributes'] ) ) {
 				$content = $this->translate_attributes( $content, $tag_name, $shortcode['attributes'] );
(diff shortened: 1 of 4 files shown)


src/media/shortcodes/class-wpml-page-builders-media-shortcodes.php
```diff
@@ -3,8 +3,8 @@
 class WPML_Page_Builders_Media_Shortcodes {
 
 	const ALL_TAGS = '\w+';
```
brucepearson: @strategio I don't think is used anymore.

strategio: Done @brucepearson (see https://github.com/OnTheGoSystems/wpml-page-builders/pull/33/commits/464fa49bf120f96cbbcc4bd346ff02a9bd9ac0d7). I removed the `*` wildcard  and the `|` which were giving the ability to match several tag names.

PR #34, opened 2018-09-25
Only stop translating the post body if the string package contains strings

brucepearson:
Only stop translating the post body if the string package contains strings - https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-4875


strategio:
It makes sense @brucepearson +1

PR #35, opened 2018-09-25
Add strings to the content parsed for media usage

strategio:
Goes with https://git.onthegosystems.com/wpml/wpml-media-translation/merge_requests/262

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlmedia-550

### DIFF
--- a/src/media/class-wpml-page-builders-media-hooks.php
+++ b/src/media/class-wpml-page-builders-media-hooks.php
@@ -20,6 +20,7 @@ class WPML_Page_Builders_Media_Hooks implements IWPML_Action {
 	}
 	public function add_hooks() {
 		add_filter( 'wmpl_pb_get_media_updaters', array( $this, 'add_media_updater' ) );
+		add_filter( 'wpml_media_content_for_media_usage', array( $this, 'add_package_strings_content' ), 10, 2 );
 	}
 	/**
 	 * @param IWPML_PB_Media_Update[] $updaters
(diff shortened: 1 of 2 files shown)


src/media/class-wpml-page-builders-media-hooks.php
```diff
@@ -32,4 +33,25 @@ public function add_media_updater( $updaters ) {
...
 		return $updaters;
 	}
+
+	/**
+	 * @param string  $content
+	 * @param WP_Post $post
+	 *
+	 * @return string
+	 */
+	public function add_package_strings_content( $content, $post ) {
```
brucepearson: @strategio I think it would be a cleaner design if this function was just a private function in `WPML_Media_Post_With_Media_Files`. I don't think there's any real need to call it via a filter.

strategio: @brucepearson, intially I did this in WPML Media, but I figured out that this is only needed for metadata page builders and this is not needed for the ones using shortcodes.

brucepearson: @strategio I thought we were handling shortcode strings in pretty much the same way as builders that have strings in metadata. If there is no simpler way or more unified way then I'm ok with your solution.

brucepearson: @strategio On second thoughts, doesn't this code also run for page builders that use shortcodes?

strategio: No @brucepearson, it runs only for metadata page builders.

For shortcodes page builders, the `` tags are already inside `$post->post_content`.

In the next task, we will look in media modules (not in strings) to find other possible media usages, and this will apply to both metadata and shortcodes page builders.

brucepearson: OK, thanks for the clarification.

PR #37, opened 2018-09-27
Escape false backward references in replacement string

strategio:
Note that we cannot use `preg_quote` on the replacement string because we would have other special chars (e.g. `.`) quoted in the final string.

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-5291

### DIFF
--- a/src/st/strategy/shortcode/class-wpml-pb-update-shortcodes-in-content.php
+++ b/src/st/strategy/shortcode/class-wpml-pb-update-shortcodes-in-content.php
@@ -91,18 +91,32 @@ class WPML_PB_Update_Shortcodes_In_Content {
 				$this->new_content = str_replace( $block, $new_block, $this->new_content );
 			} else {
 				if ( $is_attribute && $attr ) {
-					$pattern   = '/' . $attr . '=(["\'])' . preg_quote( $original, '/' ) . '(["\'])/';
-					$new_block = preg_replace( $pattern, $attr . '=${1}' . $translation . '${2}', $block );
+					$pattern     = '/' . $attr . '=(["\'])' . preg_quote( $original, '/' ) . '(["\'])/';
+					$replacement = $attr . '=${1}' . $this->escape_backward_reference_on_replacement_string( $translation ) . '${2}';
 				} else {
-					$pattern   = '/(]\s*)' . preg_quote( trim( $original ), '/' ) . '(\s*\[)/';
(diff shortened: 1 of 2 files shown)


src/st/strategy/shortcode/class-wpml-pb-update-shortcodes-in-content.php
```diff
@@ -91,20 +91,35 @@ private function replace_string_with_translation( $block, $original, $translatio
...
-					$new_block = preg_replace( $pattern, $attr . '=${1}' . $translation . '${2}', $block );
+					$pattern     = '/' . $attr . '=(["\'])' . preg_quote( $original, '/' ) . '(["\'])/';
+					$replacement = $attr . '=${1}' . $this->escape_backward_reference_on_replacement_string( $translation ) . '${2}';
+					$new_block   = preg_replace( $pattern, $replacement, $block );
 				} else {
-					$pattern   = '/(]\s*)' . preg_quote( trim( $original ), '/' ) . '(\s*\[)/';
-					$new_block = preg_replace( $pattern, '${1}' . trim( $translation ) . '${2}', $block );
+					$pattern     = '/(]\s*)' . preg_quote( trim( $original ), '/' ) . '(\s*\[)/';
+					$replacement = '${1}' . $this->escape_backward_reference_on_replacement_string( trim( $translation ) ) . '${2}';
+					$new_block   = preg_replace( $pattern, $replacement, $block );
```
andreasciamanna: I might be wrong, because I find very hard to understand the logic here, but I think this whole if-else block can be simplified a bit:

```php

$replacement = '${1}' . $this->escape_backward_reference_on_replacement_string( trim( $translation ) ) . '${2}';
if ( $is_attribute && $attr ) {
	$pattern     = '/' . $attr . '=(["\'])' . preg_quote( $original, '/' ) . '(["\'])/';
	$replacement = $attr . '=' .  $replacement;
} else {
	$pattern     = '/(]\s*)' . preg_quote( trim( $original ), '/' ) . '(\s*\[)/';
}
$new_block   = preg_replace( $pattern, $replacement, $block );
```

strategio: @andreasciamanna, I extracted one line `$new_block = preg_replace( $pattern, $replacement, $block );`, but I prefer not to go further. The translation for the content is trimmed while the one for the attribute isn't. If we trim the translation for the attribute, we probably need to change its regex pattern (and we are going beyond the scope of the initial issue).

PR #42, opened 2018-10-15
Copy original when re-saving a translation which has no strings

strategio:
Goes with https://git.onthegosystems.com/wpml/wpml-media-translation/merge_requests/269.

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-5935

### DIFF
--- a/src/st/class-wpml-pb-integration.php
+++ b/src/st/class-wpml-pb-integration.php
@@ -68,6 +68,14 @@ class WPML_PB_Integration {
 
 		$updated_packages = $this->factory->get_package_strings_resave()->from_element( $post_element );
 
+		if ( ! $updated_packages ) {
+			$this->factory->get_handle_post_body()->copy(
+				$post_element->get_id(),
+				$post_element->get_source_element()->get_id(),
+				array()
+			);
+		}
+
 		foreach ( $this->strategies as $strategy ) {
 
 			foreach ( $updated_packages as $package ) {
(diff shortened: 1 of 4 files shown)


strategio:
@brucepearson do you have any comment?

brucepearson:
@strategio Don't we have something already to handle this. Something about `built with pagebuilders` that copies the post body if there are no translated strings. Maybe it's only for shortcodes.

strategio:
@brucepearson, we do copy the content from original to translation when the translation job is saved (see https://github.com/OnTheGoSystems/wpml-page-builders/blob/develop/src/tm/class-wpml-pb-handle-post-body.php#L13). This applies to all page builders, and there's an extra logic running for metadata page builders on the action hook `wpml_pb_after_page_without_elements_post_content_copy` (see https://github.com/OnTheGoSystems/wpml-page-builders/blob/develop/src/tm/class-wpml-pb-handle-post-body.php#L35).

That's why this is needed when we re-save the translation.

src/st/class-wpml-pb-integration.php
```diff
@@ -68,6 +68,14 @@ public function resave_post_translation_in_shutdown( WPML_Post_Element $post_ele
 
 		$updated_packages = $this->factory->get_package_strings_resave()->from_element( $post_element );
 
+		if ( ! $updated_packages ) {
```
brucepearson: @strategio I would be tempted to add another hook to `WPML_PB_Handle_Post_Body` called `wpml_pb_copy_body_if_no_packages` that calls the copy function. I'll leave it to you to decide.

PR #43, opened 2018-10-16
Post count is wrongly updated.

kagg-design:
Post count is wrongly updated

Steps to reproduce:

Create post with tag 'tag1' and some PB shortcode like `[vc_text_separator]`
Duplicate it to secondary language (let say `sr`).
Post count is wrong, because duplicate got wrong tag. Not `tag1-sr`, but simply `tag1`.

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-5938

-----

First solution:

Check if post content has no PB tags and do not update the post.

Final solution:

Workaround in wp_update_post() call, as it is the bug in WP core.

### DIFF
--- a/src/media/shortcodes/class-wpml-page-builders-media-shortcodes-update.php
+++ b/src/media/shortcodes/class-wpml-page-builders-media-shortcodes-update.php
@@ -25,10 +25,6 @@ class WPML_Page_Builders_Media_Shortcodes_Update implements IWPML_PB_Media_Updat
 	 * @param WP_Post $post
 	 */
 	public function translate( $post ) {
-		if ( $this->has_no_shortcode( $post->post_content ) ) {
-			return;
-		}
-
 		if ( ! $this->media_shortcodes->has_media_shortcode( $post->post_content ) ) {
 			return;
 		}
@@ -45,10 +41,15 @@ class WPML_Page_Builders_Media_Shortcodes_Update implements IWPML_PB_Media_Updat
 
 		$this->media_usage->update( $element->get_source_element()->get_id() );
 
-		wp_update_post( $post );
-	}
-
-	private function has_no_shortcode( $content ) {
-		return strpos( $content, '[' ) === false;
+		// wp_update_post() can modify post tag. Code below sends tags by IDs to prevent this.
+		// wpmlcore-5947
+		// https://core.trac.wordpress.org/ticket/45121
+		$tag_ids = wp_get_post_tags( $post->ID, array( 'fields' => 'ids' ) );
+		$postarr = array(
+			'ID'           => $post->ID,
+			'post_content' => $post->post_content,
+			'tags_input'   => $tag_ids,
+		);
+		wp_update_post( $postarr );
 	}
 }
--- a/src/media/shortcodes/class-wpml-page-builders-media-shortcodes-update.php
+++ b/src/media/shortcodes/class-wpml-page-builders-media-shortcodes-update.php
@@ -29,6 +29,10 @@ class WPML_Page_Builders_Media_Shortcodes_Update implements IWPML_PB_Media_Updat
 			return;
 		}
 
+		if ( ! $this->media_shortcodes->has_media_shortcode( $post->post_content ) ) {
+			return;
+		}
+
 		$element = $this->element_factory->create_post( $post->ID );
 
 		if ( ! $element->get_source_language_code() ) {
--- a/src/media/shortcodes/class-wpml-page-builders-media-shortcodes.php
+++ b/src/media/shortcodes/class-wpml-page-builders-media-shortcodes.php
@@ -45,6 +45,24 @@ class WPML_Page_Builders_Media_Shortcodes {
 	}
 
 	/**
+	 * @param string $content
+	 *
+	 * @return bool
+	 */
+	public function has_media_shortcode( $content ) {
+		foreach ( $this->config as $shortcode ) {
+			$shortcode = $this->sanitize_shortcode( $shortcode );
+			$tag_name  = isset( $shortcode['tag']['name'] ) ? $shortcode['tag']['name'] : '';
+
+			if ( $tag_name && false !== strpos( $content, '[' . $tag_name ) ) {
+				return true;
+			}
+		}
+
+		return false;
+	}
+
+	/**
 	 * @param array $shortcode
 	 *
 	 * @return array
--- a/src/media/shortcodes/class-wpml-page-builders-media-shortcodes.php
+++ b/src/media/shortcodes/class-wpml-page-builders-media-shortcodes.php
@@ -50,6 +50,10 @@ class WPML_Page_Builders_Media_Shortcodes {
 	 * @return bool
 	 */
 	public function has_media_shortcode( $content ) {
+		if ( false === strpos( $content, '[' ) ) {
+			return false;
+		};
+
 		foreach ( $this->config as $shortcode ) {
 			$shortcode = $this->sanitize_shortcode( $shortcode );
 			$tag_name  = isset( $shortcode['tag']['name'] ) ? $shortcode['tag']['name'] : '';
--- a/tests/phpunit/tests/media/shortcodes/test-wpml-page-builders-media-shortcodes-update.php
+++ b/tests/phpunit/tests/media/shortcodes/test-wpml-page-builders-media-shortcodes-update.php
@@ -11,20 +11,6 @@ class Test_WPML_Page_Builders_Media_Shortcodes_Update extends \OTGS\PHPUnit\Tool
 	/**
 	 * @test
 	 */
-	public function it_should_not_translate_if_content_has_no_shortcode() {
-		$post = $this->get_post( 'Some content with no shortcode' );
-
-		\WP_Mock::userFunction( 'wp_update_post', array(
-			'times' => 0,
-		));
-
-		$subject = $this->get_subject();
-		$subject->translate( $post );
-	}
-
-	/**
-	 * @test
-	 */
 	public function it_should_not_translate_if_post_is_source() {
 		$post = $this->get_post( '[shortcodeA]Hello[/shortcodeA]' );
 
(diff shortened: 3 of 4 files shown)


andreasciamanna:
I agree with Pierre's concerns.

The main commit message should contain an explanation of the fix and the same applies to the PR description (which anyway will be automatically created with the commit message, in most of the cases).

When we go back to the commit history, we should know why a change was made.  
Just referencing the ticket is not enough, especially if that ticket is also lacking an explanation.

kagg-design:
Problem is much more deeper and not related to Multilingual Tools plugin.

Steps to reproduce:
1. Create post with tag 'tag1' and some PB shortcode like `[vc_text_separator]`
2. Duplicate it to secondary language (let say `sr`).

Post count is wrong, because duplicate got wrong tag. Not `tag1-sr`, but simply `sr`.

This caused by the following lines in `wp-includes/post.php`, `wp_insert_post()` function:

	if ( isset( $postarr['tags_input'] ) && is_object_in_taxonomy( $post_type, 'post_tag' ) ) {
		wp_set_post_tags( $post_ID, $postarr['tags_input'] );
	}

If `tags_input` is not set in a call to `wp_update_post()`, then it is taken by name. Name is the same: `tag1`, but ids and slugs are different (`tag1` and `tag1-sr`).

To avoid this problem, we need to call `wp_update_post()` with array of assotiated tag ids. This is done in the commit.

Manual tests after 5 hours of debug are OK :) Some CC tests should be added to check post count in admin.

strategio:
Thanks @kagg-design! Honestly, it sounds strange to me that we have to fetch the post tags before to save the post. We are running a similar logic for strings and I don't see why it would not happen there. I will give it a try tomorrow, now my mind is not responsive anymore :D

kagg-design:
@strategio Please look at two snapshots,. Just one step in the debugger. This happens when we call `wp_update_post($post)` without my modifications.

kagg-design:
@strategio here is the simple plugin to test: https://github.com/kagg-design/update-post-bug

Activate it on clean WP. Plugin creates two tags with slugs `wp_update_post_tag_1` and `wp_update_post_tag_2` and the same name `wp_update_post_tag`.

Go to `/wpup-test/` page of your site. You will see output like that

wp_get_post_tags() Before wp_update_post()

array (size=1)
  0 => 
    object(WP_Term)[312]
      public 'term_id' => int 39
      public 'name' => string 'wp_update_post_tag' (length=18)
      public 'slug' => string 'wp_update_post_tag_2' (length=20)
      public 'term_group' => int 0
      public 'term_taxonomy_id' => int 39
      public 'taxonomy' => string 'post_tag' (length=8)
      public 'description' => string '' (length=0)
      public 'parent' => int 0
      public 'count' => int 1
      public 'filter' => string 'raw' (length=3)

wp_get_post_tags() After wp_update_post()

array (size=1)
  0 => 
    object(WP_Term)[314]
      public 'term_id' => int 38
      public 'name' => string 'wp_update_post_tag' (length=18)
      public 'slug' => string 'wp_update_post_tag_1' (length=20)
      public 'term_group' => int 0
      public 'term_taxonomy_id' => int 38
      public 'taxonomy' => string 'post_tag' (length=8)
      public 'description' => string '' (length=0)
      public 'parent' => int 0
      public 'count' => int 1
      public 'filter' => string 'raw' (length=3)

And that's nice. tag attached to the post has changed.

src/media/shortcodes/class-wpml-page-builders-media-shortcodes.php
```diff
@@ -44,6 +44,24 @@ public function translate( $content )  {
 		return $content;
 	}
 
+	/**
+	 * @param string $content
+	 *
+	 * @return bool
+	 */
+	public function has_media_shortcode( $content ) {
```
strategio: @kagg-design, I suggest moving the code from https://github.com/OnTheGoSystems/wpml-page-builders/blob/4528acc9c636bd8db069dfda16c44fec5c5fd5a5/src/media/shortcodes/class-wpml-page-builders-media-shortcodes-update.php#L57 to the beginning of this method.

So basically, before to check if one shortcode is matching, we will check if there is any shortcode in the content.

src/media/shortcodes/class-wpml-page-builders-media-shortcodes-update.php
```diff
@@ -41,7 +45,13 @@ public function translate( $post ) {
 
 		$this->media_usage->update( $element->get_source_element()->get_id() );
 
-		wp_update_post( $post );
+		$tag_ids = wp_get_post_tags( $post->ID, array( 'fields' => 'ids' ) );
```
strategio: Please add a comment to explain the reason for this code otherwise it does not make sense (include a link to the related WP Trac ticket and YT ticket).

src/media/shortcodes/class-wpml-page-builders-media-shortcodes.php
kagg-design: Done.

src/media/shortcodes/class-wpml-page-builders-media-shortcodes-update.php
kagg-design: Done.

kagg-design:
@strategio Please review again.

tests/phpunit/tests/media/shortcodes/test-wpml-page-builders-media-shortcodes-update.php
```diff
@@ -8,20 +8,6 @@ class Test_WPML_Page_Builders_Media_Shortcodes_Update extends \OTGS\PHPUnit\Tool
 	const ORIGINAL_POST_ID   = 9;
 	const TRANSLATED_POST_ID = 57;
 
-	/**
-	 * @test
-	 */
-	public function it_should_not_translate_if_content_has_no_shortcode() {
```
strategio: @kagg-design, you can keep this test but you need to configure `WPML_Page_Builders_Media_Shortcodes`'s mock to return `false` for the method `has_media_shortcode`.

tests/phpunit/tests/media/shortcodes/test-wpml-page-builders-media-shortcodes.php
```diff
@@ -141,6 +141,181 @@ public function it_should_translate() {
...
+	 * @test
+	 * @group wpmlpb-166
+	 */
+	public function it_should_return_true_when_content_has_media_shortcodes() {
+		$original_url_1 = 'http://example.org/dog.jpg';
+		$original_url_2 = 'http://example.org/england.jpg';
+		$original_id_1  = 2;
+		$original_id_2  = 7;
+
+		$content = '
```
strategio: @kagg-design, to make the test more readable, you can reduce the number of shortcodes in the config and also reduce the content:

```php
$content = '[et_pb_gallery _builder_version="3.12" gallery_ids="1,2" /]';

$config = array(
     array(
				'tag'        => array( 'name' => 'et_pb_gallery' ),
				'attributes' => array(
					'gallery_ids' => array( 'type' => WPML_Page_Builders_Media_Shortcodes::TYPE_IDS ),
				),
			),
);
```

tests/phpunit/tests/media/shortcodes/test-wpml-page-builders-media-shortcodes.php
```diff
@@ -141,6 +141,181 @@ public function it_should_translate() {
...
+	}
+
+	/**
+	 * @test
+	 * @group wpmlpb-166
+	 */
+	public function it_should_return_false_when_content_has_NO_media_shortcodes() {
+		$content = '[sr]';
+
+		$config = array(
```
strategio: Same for this test, you can have only two shortcodes (not matching).

tests/phpunit/tests/media/shortcodes/test-wpml-page-builders-media-shortcodes-update.php
kagg-design: As code is moved to `WPML_Page_Builders_Media_Shortcodes`, so test is moved to `Test_WPML_Page_Builders_Media_Shortcodes` and named `it_should_return_false_when_content_has_no_shortcode()`

tests/phpunit/tests/media/shortcodes/test-wpml-page-builders-media-shortcodes.php
kagg-design: Done

tests/phpunit/tests/media/shortcodes/test-wpml-page-builders-media-shortcodes.php
kagg-design: Done

kagg-design:
@pierre please review, tests are simplified according your notes

tests/phpunit/tests/media/shortcodes/test-wpml-page-builders-media-shortcodes-update.php
kagg-design: @strategio test added, please check

tests/phpunit/tests/media/shortcodes/test-wpml-page-builders-media-shortcodes-update.php
```diff
@@ -27,6 +27,53 @@ public function it_should_not_translate_if_post_is_source() {
...
+		$source_element->method( 'get_id' )->willReturn( $original_post_id );
+
+		$element = $this->get_element( $target_lang, $source_lang );
+		$element->method( 'get_source_element' )->willReturn( $source_element );
+
+		$factory = $this->get_element_factory();
+		$factory->method( 'create_post' )->with( $post->ID )->willReturn( $element );
+
+		$media_shortcodes = $this->getMockBuilder( 'WPML_Page_Builders_Media_Shortcodes' )
+		                   ->setMethods( array( 'set_target_lang', 'set_source_lang', 'translate', 'has_media_shortcode' ) )
```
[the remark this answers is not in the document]
kagg-design: @strategio I left them for a reason. I tested this tests with commended out `if ( ! $this->media_shortcodes->has_media_shortcode( $post->post_content ) ) {` And made test running but with wrong assertion related to count of wp_update_post()

Let us leave it as is.

tests/phpunit/tests/media/shortcodes/test-wpml-page-builders-media-shortcodes-update.php
```diff
@@ -27,6 +27,53 @@ public function it_should_not_translate_if_post_is_source() {
...
+		);
+
+		$source_element = $this->get_element( $source_lang, null );
+		$source_element->method( 'get_id' )->willReturn( $original_post_id );
+
+		$element = $this->get_element( $target_lang, $source_lang );
+		$element->method( 'get_source_element' )->willReturn( $source_element );
+
+		$factory = $this->get_element_factory();
+		$factory->method( 'create_post' )->with( $post->ID )->willReturn( $element );
```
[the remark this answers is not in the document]
kagg-design: @strategio pls see above...

tests/phpunit/tests/media/shortcodes/test-wpml-page-builders-media-shortcodes-update.php
```diff
@@ -27,6 +27,53 @@ public function it_should_not_translate_if_post_is_source() {
...
+			'wp_get_post_tags',
+			array(
+				'times' => 0,
+			)
+		);
+
+		\WP_Mock::userFunction(
+			'wp_update_post',
+			array(
+				'times' => 0,
```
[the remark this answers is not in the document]
kagg-design: @strategio yes, I understand it

dimitrismitsis:
This pull request was a bit messed up.  
Commits like this one: Merge branch 'develop' into wpmlpb-166 are wrong.  
Probably done due to something wrong while rebasing.  
These commits are wrong since we apply fast-forward merging to all projects.  FYI @kagg-design  
Merged manually to `develop`

dimitrismitsis:
This pull request was a bit messed up.  
Commits like this one: Merge branch 'develop' into wpmlpb-166 are wrong.  
Probably done due to something wrong while rebasing.  
These commits are wrong since we apply fast-forward merging to all projects.  FYI @kagg-design  
Merged manually to `develop`

PR #44, opened 2018-10-22
Prevent adding new translations when a string is registered

strategio:
In some cases, WPML PB updates the string translations in all languages
(e.g. a `LINK` string) and this is forcing to re-save all the post
translations in the shutdown.

This was prevented in `do_shutdown_action` by removing the action hook
on `icl_st_add_string_translation`, but not for
`cleanup_strings_after_translation_completed`.

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-5872


strategio:
@OnTheGoSystems/wpml, one more review please.

PR #47, opened 2018-11-08
Check if table icl_string_packages exists before querying it

adrianosferreira:
Internally, TM and Core are dependencies of ST, not having any of them are causing the database schema update (which creates icl_string_packages) to not be fired thus throws a database error in \WPML_PB_Integration::post_has_strings.

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-6021


adrianosferreira:

PR #48, opened 2018-11-23
Extract regular strings mixed with shortcodes inside shortcode content

strategio:
The general idea is to wrap those strings in a generic shortcode. This will allow to register the related strings, and to replace it when the content is saved (note that this wrapper is only added for our process and does not persist in the DB entry).

Example (consider the content below as the content of another shortcode):

```
Some text
[shortcode_A title="My title"]
```

Will become:

```
[wpml_string_wrapper]Some text[wpml_string_wrapper]
[shortcode_A title="My title"]
```

In the translation editor, we will have 2 fields ("Some text" and "My title") instead of one field with the whole content.

The hardest part is to wrap the strings because we have a lot of possible cases. To achieve this, I created specialized class `WPML_PB_Shortcode_Content_Wrapper` which proceed as follow:
1. Split the content into an array of chars (so we can analyze it properly).
2. Find all the valid shortcodes inside the content (we want to exclude those parts for the rest of the analysis).
3. Find valid unwrapped strings and set the positions where we will insert a wrapper (open/close).
4. Insert the wrappers.

It's a partial shortcode parser which will just fulfill our needs.

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmltm-2970

**Note:** I could not find a shortcode parser in existing public packages. It's either not compatible with PHP 5.2.4 or not appropriate for what we need.


src/st/strategy/shortcode/class-wpml-pb-update-shortcodes-in-content.php
```diff
@@ -94,13 +95,30 @@ private function replace_string_with_translation( $block, $original, $translatio
...
 				} else {
+					$has_wrapper = false !== strpos( $new_block, '[' . WPML_PB_Shortcode_Content_Wrapper::WRAPPER_SHORTCODE_NAME . ']' );
 					$pattern     = '/(]\s*)' . preg_quote( trim( $original ), '/' ) . '(\s*\[)/';
 					$replacement = '${1}' . $this->escape_backward_reference_on_replacement_string( trim( $translation ) ) . '${2}';
 				}
 
-				$new_block         = preg_replace( $pattern, $replacement, $block );
-				$replacement       = $this->escape_backward_reference_on_replacement_string( $new_block );
-				$this->new_content = preg_replace( '/' . preg_quote( $block, '/' ) . '/', $replacement, $this->new_content, 1 );
+				$new_block = preg_replace( $pattern, $replacement, $block );
```
brucepearson: @strategio This code seems to complicated to me. Why can we just run code to remove the wrappers by default?

```
$new_block         = preg_replace( $pattern, $replacement, $block );
$new_block         = $this->remove_wrappers( $new_block );
$replacement       = $this->escape_backward_reference_on_replacement_string( $new_block );
$this->new_content = preg_replace( '/' . preg_quote( $block, '/' ) . '/', $replacement, $this->new_content, 1 );
```

There's also a bug in the code where the code remove wrappers doesn't run if `is_string_too_long_for_regex` return true.

src/st/strategy/shortcode/class-wpml-pb-update-shortcodes-in-content.php
```diff
@@ -82,25 +82,35 @@ private function update_shortcode_attributes( $shortcode_data ) {
...
-				$new_block         = preg_replace( $pattern, $replacement, $block );
-				$replacement       = $this->escape_backward_reference_on_replacement_string( $new_block );
-				$this->new_content = preg_replace( '/' . preg_quote( $block, '/' ) . '/', $replacement, $this->new_content, 1 );
+				$new_block   = preg_replace( $pattern, $replacement, $block );
+				$block       = $this->remove_wrappers( $block );
+				$new_block   = $this->remove_wrappers( $new_block );
+				$replacement = $this->escape_backward_reference_on_replacement_string( $new_block );
+
+				if ( $has_wrapper ) {
+					$this->new_content = preg_replace( '/(]\s*)' . preg_quote( $block, '/' ) . '(\s*\[)/', '${1}' . $replacement . '${2}', $this->new_content, 1 );
```
brucepearson: @strategio Why is this special case here? If we have already removed the wrappers in $block and $new_block I don't understand why this special case.

strategio: @brucepearson, since we don't have any delimiters for the string, this will ensure that the string we are replacing is (at least) inside a shortcode content (between `]` and `[`).

Otherwise, we could be replacing an attribute name (e.g. the string "title" could replace the attribute name `title`).

I will try to reflect this need in the tests.

strategio: Added a new test @brucepearson (see https://github.com/OnTheGoSystems/wpml-page-builders/pull/48/files#diff-81155f1140cd32496c3943c384191c0bR340). This test is failing if we go through the usual case.

tests/phpunit/tests/st/strategy/shortcode/test-wpml-pb-shortcode-content-wrapper.php
```diff
@@ -0,0 +1,132 @@
...
+			'shortcode_nested',
+			'shortcode_A',
+		);
+
+		$subject = new WPML_PB_Shortcode_Content_Wrapper( $content, $valid_shortcodes );
+		$this->assertEquals( $expected_content, $subject->get_wrapped_content() );
+	}
+
+	public function dp_get_wrapped_content() {
+		return array(
```
brucepearson: @strategio This is very hard to read and understand. To add extra test cases is messy and makes in harder to understand.

Can we change this to load source html and parsed html from files in a fixtures directory under the tests?

eg.

basic_test.html file
```html
This is the main text.

[av_button label='Click me' link='manually,http://google.com']
```

basic_test_parsed.html file
```html
[wpml_string_wrapper]This is the main text.[/wpml_string_wrapper]

[av_button label='Click me' link='manually,http://google.com']
```

src/st/strategy/shortcode/class-wpml-pb-update-shortcodes-in-content.php
brucepearson: I'm still struggling to easily understand the regex here. Can you move it to a private function with a meaning full name.
I still do understand don't need this special handling when we don't have wrappers. Don't we have the same problem?

strategio: @brucepearson, when we translate a usual shortcode content, we first translate it inside its block:

`[shortcode_A]My text[/shortcode_A]` will become `[shortcode_A]FR My text[/shortcode_A]`.

Then, we replace the whole block inside the whole content.

With wrapped content, we can't do this because the wrappers do not exist in the actual content (we never inject the wrappers in the actual content). So we replace a simple string (with no delimiters) inside the whole content.

To limit possible wrong replacements (explained in https://github.com/OnTheGoSystems/wpml-page-builders/pull/48#discussion_r236255777), we make sure that this replacement is done inside a shortcode content.

I moved the replacement line into a private method `replace_content_without_delimiters`.

tests/phpunit/tests/st/strategy/shortcode/test-wpml-pb-shortcode-content-wrapper.php
strategio: Done @brucepearson

brucepearson:
@strategio I'm still trying to understand how all this works. I tried a simple test with the following content:

```
This text is outside the block
<p>[et_pb_section admin_label="section"][et_pb_row admin_label="row"][et_pb_column type="4_4"][et_pb_text admin_label="Text" background_layout="light" text_orientation="left" use_border_color="off" border_color="#ffffff" border_style="solid"]</p><p>This is in a block</p><p>[/et_pb_text][/et_pb_column][/et_pb_row][/et_pb_section]</p>
```

Only the block et_pb_text was registered for translation.

tests/phpunit/tests/st/strategy/shortcode/fixtures/content-wrapper/empty-strings-between-2-shortcodes-wrapped.html
```diff
@@ -0,0 +1,3 @@
+[shortcode_A]	[av_notification title='Note']
```
brucepearson: @strategio What is `shortcode_A`? I assume it's an unknown shortcode. Mabye `[unknown_shortcode_A]` would be clearer.

strategio: No @brucepearson, the `shortcode_A` is a valid one. In this content, we don't have any regular text to translate (and we don't want to handle the whitespaces, tabs, ...).

strategio: If `shortcode_A` was not a registered shortcode, we would wrap it.

tests/phpunit/tests/st/strategy/shortcode/test-wpml-pb-shortcode-content-wrapper.php
```diff
@@ -0,0 +1,55 @@
...
+			'basic text'                                       => $this->get_fixture_contents( 'basic-text' ),
+			'with open and closing shortcode'                  => $this->get_fixture_contents( 'with-open-and-closing-shortcode' ),
+			'with 2 shortcode side by side'                    => $this->get_fixture_contents( 'with-2-shortcodes-side-by-side' ),
+			'with nested shortcodes'                           => $this->get_fixture_contents( 'with-nested-shortcodes' ),
+			'start and end with shortcodes'                    => $this->get_fixture_contents( 'start-and-end-with-shortcodes' ),
+			'"empty" strings between 2 shortcodes (no change)' => $this->get_fixture_contents( 'empty-strings-between-2-shortcodes' ),
+			'with invalid shortcodes (should wrap it)'         => $this->get_fixture_contents( 'with-invalid-shortcodes' ),
+			'with brackets in shortcode attributes'            => $this->get_fixture_contents( 'with-brackets-in-shortcode-attributes' ),
+			'with multibyte chars'                             => $this->get_fixture_contents( 'with-multibyte-chars' ),
+		);
```
brucepearson: @strategio Why not just `glob` the directory to find the files. Then it's easier to add more cases by just adding new `xxx.html` file and `xxx-wrapped.html` without having to adjust the php.

strategio:
@OnTheGoSystems/wpml one more review please.

strategio:
@OnTheGoSystems/wpml please one more review!

@brucepearson could you re-approve this?

strategio:
@OnTheGoSystems/wpml please review :)

PR #50, opened 2018-12-05
WIP Force body translation when a PB page is updated with the native editor

strategio:
If the content was previously translated with a job, this will prevent the body to be overwritten with the PB strings.

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-6120


strategio:
Closing as @brucepearson suggested a better approach in WPML Media.

PR #55, opened 2019-01-02
Add code from PR31 which is missing

strategio:
This is possibly due to a rebase conflict which was not properly
resolved (see https://github.com/OnTheGoSystems/wpml-page-builders/pull/31/files).

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-5811


strategio:
@adrianosferreira, could you please give it a look?

PR #56, opened 2019-01-03
Always run the shortcode strategy at last

strategio:
In the API strategy, we usually don't care of the post content but
Cornerstone is using both shortcodes (legacy blocks) and metadata.

When the API strategy saves the content, it also copies the post_content
from original to translation, and this was overwritting the translations
applied to the shortcodes.

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-6208

### DIFF
--- a/src/st/class-wpml-pb-loader.php
+++ b/src/st/class-wpml-pb-loader.php
@@ -12,6 +12,17 @@ class WPML_PB_Loader {
 
 		$page_builder_strategies = array();
 
+		/**
+		 * This filter hook provide the API page builders names that need to be supported.
+		 *
+		 * For each PB name, we will create a dedicated strategy and a proper string package namespace.
+		 *
+		 * It's called in 2 places:
+		 * - `WPML_Page_Builders_Integration` for external plugins
+		 * - `WPML_Gutenberg_Integration` for WordPress Core block editor
+		 *
+		 * @param string[] Required plugin names (e.g. `Beaver Builder`, `Gutenberg`)
+		 */
 		$required = apply_filters( 'wpml_page_builder_support_required', array() );
 		foreach ( $required as $plugin ) {
 			$page_builder_strategies[] = new WPML_PB_API_Hooks_Strategy( $plugin );
--- a/src/st/class-wpml-pb-loader.php
+++ b/src/st/class-wpml-pb-loader.php
@@ -12,6 +12,11 @@ class WPML_PB_Loader {
 
 		$page_builder_strategies = array();
 
+		$required = apply_filters( 'wpml_page_builder_support_required', array() );
+		foreach ( $required as $plugin ) {
+			$page_builder_strategies[] = new WPML_PB_API_Hooks_Strategy( $plugin );
+		}
+
 		$page_builder_config_import = new WPML_PB_Config_Import_Shortcode( $st_settings );
 		$page_builder_config_import->add_hooks();
 		if ( $page_builder_config_import->has_settings() ) {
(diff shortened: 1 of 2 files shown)


adrianosferreira:
@strategio nice, looks good to me. +1

src/st/class-wpml-pb-loader.php
```diff
@@ -12,6 +12,11 @@ public function __construct(
 
 		$page_builder_strategies = array();
 
+		$required = apply_filters( 'wpml_page_builder_support_required', array() );
```
andreasciamanna: @strategio can you please add PHPDoc to the filter?

It would help to know where the corresponding `add_filter` is too: I can't find it anywhere in our code.

strategio: Done @andreasciamanna! We use it in 2 places:
- `WPML_Page_Builders_Integration` which is used by Elementor and BB.
- `WPML_Gutenberg_Integration`

strategio: @andreasciamanna, I know the link you shared. But what's really missing? I skipped the `@since` attribute because I have not accurrate idea when it was added (or I could put `Unknown`).

Should we really mention where `add_filter` is called? This is something that may change and the comment will not be updated.

andreasciamanna: - The description `This filters the API page builder plugins that need to be supported` doesn't say much, does it?
I would write the actual reason for calling this filter.
- If the argument is not needed or ignored, that must be specified.
- `@since` can be skipped

> Should we really mention where add_filter is called? This is something that may change and the comment will not be updated.

The fact that it's called from two different places does not look right to me. I think it highlights a design flaw we should address in the future.  
I guess you can at least mention the two places where it's used as an example of implementation.

strategio: @andreasciamanna, I improved the PHPDoc. Please check again!

I am not sure why we have 2 classes filtering this hook. My guess is that we started with 3rd part plugins which all have a common pattern (content store in metadata). When @brucepearson started to implement the support for Gutenberg, he found it in the mid way between shortcode and metadata page builders (Gutenberg store the content in the post body). So instead of changing the existing API `WPML_Page_Builders_Integration`, he decided to create a custom one for Gutenberg.

src/st/class-wpml-pb-loader.php
```diff
@@ -12,6 +12,22 @@ public function __construct(
...
+		/**
+		 * This filter hook provide the API page builders names that need to be supported.
+		 *
+		 * For each PB name, we will create a dedicated strategy and a proper string package namespace.
+		 *
+		 * It's called in 2 places:
+		 * - `WPML_Page_Builders_Integration` for external plugins
+		 * - `WPML_Gutenberg_Integration` for WordPress Core block editor
+		 *
+		 * @param string[] Required plugin names (e.g. `Beaver Builder`, `Gutenberg`)
```
andreasciamanna: @kkarpieszuk this parameter is never passed: why you are documenting it as "required"?

I would change it into `string[] Not used`.

Then, in the description, you can add something like "it returns the plugin names (e.g. `Beaver Builder`, `Gutenberg`)".

strategio: @andreasciamanna, in my understanding the first parameter is the value we actually want to filter (an empty `array` by default).

I used "Required" as an adjective of "plugin" (i.e. "The name of the plugins that are required".

I am a bit confused. Maybe it would be faster if you paste here the PHPDoc that you have in mind.

PR #58, opened 2019-01-10
Check if original post is set before copying the content of it

adrianosferreira:
In certain situations `wpml_translation_job_saved` can receive `false` value as first argument, so no post will be found and cause an error.

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-6222


adrianosferreira:

PR #60, opened 2019-01-22
Include $matches argument in preg_match_all in order to be fully compatible with PHP 5.3.x

adrianosferreira:
This parameter became optional from PHP 5.4.0.

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-6269


adrianosferreira:

PR #61, opened 2019-02-04
Mark headings in XLIFFs so that ATE can indicate that these fields have SEO significance.

kagg-design:
Changes in wpml-page-builders to support task of the ticket.

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmltm-3081

### DIFF
--- a/src/tm/class-wpml-tm-page-builders-field-wrapper.php
+++ b/src/tm/class-wpml-tm-page-builders-field-wrapper.php
@@ -94,20 +94,14 @@ class WPML_TM_Page_Builders_Field_Wrapper {
 	}
 
 	/**
-	 * @param string $string_name
+	 * Get string wrap.
+	 *
+	 * @param stdClass $string WPML string.
 	 *
 	 * @return string
 	 */
-	public static function get_wrap( $string_name ) {
-		$heading = '';
-		if ( strpos( $string_name, '-heading-' ) ) {
-			$name_arr = explode( '-', $string_name );
-			if ( isset( $name_arr[3] ) ) {
-				$heading = $name_arr[3];
-			}
-		}
-
-		return $heading;
+	public static function get_wrap( $string ) {
+		return isset( $string->wrap ) ? $string->wrap : '';
 	}
 
 	/**
@@ -117,11 +111,7 @@ class WPML_TM_Page_Builders_Field_Wrapper {
 	 * @return string
 	 */
 	public static function generate_field_slug( $package_id, $string ) {
-		$wrap = self::get_wrap( $string->name );
-		if ( '' !== $wrap ) {
-			$wrap = '-' . $wrap;
-		}
-		return self::SLUG_BASE . $package_id . '-' . $string->id . $wrap;
+		return self::SLUG_BASE . $package_id . '-' . $string->id;
 	}
 
 	/**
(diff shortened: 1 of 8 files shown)


kagg-design:
@OnTheGoSystems/wpml Please review

src/tm/class-wpml-tm-page-builders-field-wrapper.php
```diff
@@ -93,14 +93,35 @@ public function get_string_type() {
...
+	 * @param string $string_name
+	 *
+	 * @return string
+	 */
+	public static function get_wrap( $string_name ) {
+		$heading = '';
+		if ( strpos( $string_name, '-heading-' ) ) {
+			$name_arr = explode( '-', $string_name );
+			if ( isset( $name_arr[3] ) ) {
+				$heading = $name_arr[3];
```
andreasciamanna: @kagg-design are you sure you want the fourth element of the array?

This method is not covered by tests and I can't see what a possible value of `$string_name´ could be.

Can you please add tests coverage, so you I'll also get an answer to my question?

strategio:
@kagg-design, your solution should work but I see 2 drawbacks:
- The HTML tag that wraps the string is a sort of metadata for the string. How will we deal if some day we have another kind of metadata for strings?
- The structure of the string name is uncertain, it could already have some extra `-` in it, and we would extract wrong information.

IMHO, you should evaluate the insertion of a new column in `icl_strings`. It might not be the perfect solution, but it's cleaner and more extensible I think. This solution could also cover non-PB strings (if not covered immediately, it could be done in the future).

What do you think?

src/tm/class-wpml-tm-page-builders-field-wrapper.php
kagg-design: @andreasciamanna test added. We need the 4th element here, as I see no other way to send heading type from Page Builder (like Elementor, for instance), to further processing. Or, as Pierre offers, we need to add one more field to icl_strings table.

andreasciamanna: @Igor-WPML 

I'm still a bit confused.

How exactly something like `title-heading-8e0908e-h2` is created? 

Is it the page builder (e.g. Elementor) creating this string, or is this library (wpml-page-builders)?

kagg-design:
@Pierre I agree in principle with your considerations. However, I do not think it could be any additional '-' in string name, including '-heading-', as 'heading' is provided by Elementor. Maybe I will see some problem further with other page builders, and in that case will return to the idea of adding a new column to icl_strings table.

src/tm/class-wpml-tm-page-builders-field-wrapper.php
kagg-design: @andreasciamanna it is created by wpml-page-builders-elementor: https://github.com/OnTheGoSystems/wpml-page-builders-elementor/pull/63

andreasciamanna: @Igor-WPML you should probably address Bruce's comment on OnTheGoSystems/wpml-page-builders-elementor#63

kagg-design:
@andreasciamanna @strategio I close this PR, as I did mistake in the branch name. Instead, this one I have opened: https://github.com/OnTheGoSystems/wpml-page-builders/pull/63

I the new PR I followed Pierre's suggestion to add column to strings table.

PR #64, opened 2019-02-07
Set string location when registered for API page builder

strategio:
https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-6331


strategio:
@OnTheGoSystems/wpml please review!

PR #65, opened 2019-03-15
PHP warning and notice messages in CTE after deletion of the string package.

kagg-design:
https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlst-1859


kagg-design:
@OnTheGoSystems/wpml Please review and take a look at first commit only.

PR #67, opened 2019-05-23
Prevent loading Page Builders integration if TM is not active

strategio:
I know it might not be needed anymore soon, but I suggest to fix this anyway.

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-6640

### DIFF
--- a/src/st/class-wpml-page-builders-app.php
+++ b/src/st/class-wpml-page-builders-app.php
@@ -25,7 +25,7 @@ class WPML_Page_Builders_App {
 	}
 
 	public function load_integration() {
-		if ( ! class_exists( 'WPML_ST_Package_Factory' ) ) {
+		if ( ! class_exists( 'WPML_ST_Package_Factory' ) || ! defined( 'WPML_TM_VERSION' ) ) {
 			return;
 		}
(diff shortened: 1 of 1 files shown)


strategio:
@OnTheGoSystems/wpml review please.

src/st/class-wpml-page-builders-app.php
```diff
@@ -25,7 +25,7 @@ public function add_hooks() {
 	}
 
 	public function load_integration() {
-		if ( ! class_exists( 'WPML_ST_Package_Factory' ) ) {
+		if ( ! class_exists( 'WPML_ST_Package_Factory' ) || ! defined( 'WPML_TM_VERSION' ) ) {
```
brucepearson: @strategio I don't think this is the right fix. We should still be able to create string packages without TM. The check you should make is where we are integrating the TM functions themselves.

strategio: @brucepearson, I think I see your point. In the case a user works with Core and ST, create some posts. Later activate TM and tries to translate immediately...

OK, I will close this one!

PR #70, opened 2019-08-13
Escape slashes in post arrays before to call `wp_update_post`

strategio:
https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-6774

### DIFF
--- a/src/media/shortcodes/class-wpml-page-builders-media-shortcodes-update.php
+++ b/src/media/shortcodes/class-wpml-page-builders-media-shortcodes-update.php
@@ -50,6 +50,6 @@ class WPML_Page_Builders_Media_Shortcodes_Update implements IWPML_PB_Media_Updat
 			'post_content' => $post->post_content,
 			'tags_input'   => $tag_ids,
 		);
-		wp_update_post( $postarr );
+		wpml_update_escaped_post( $postarr );
 	}
 }
--- a/.scrutinizer.yml
+++ b/.scrutinizer.yml
@@ -5,8 +5,8 @@ checks:
         duplication: true
 filter:
     excluded_paths:
-    - "tests/"
-    - "vendor/"
+        - "tests/"
+        - "vendor/"
 coding_style:
     php:
         indentation:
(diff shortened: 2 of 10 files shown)


strategio:
@OnTheGoSystems/wpml please review (I am checking why tests are failing in Scrutinizer, looks like a false alarm).

src/media/shortcodes/class-wpml-page-builders-media-shortcodes-update.php
```diff
@@ -50,6 +50,6 @@ public function translate( $post ) {
 			'post_content' => $post->post_content,
 			'tags_input'   => $tag_ids,
 		);
-		wp_update_post( $postarr );
+		wp_update_post( wp_slash( $postarr ) );
```
andreasciamanna: @strategio WordPress already calls `wp_slash` inside `wp_update_post` as long as you pass an object.
You even wrote that in the ticket (*).
So why not passing an object, instead of calling an extra function which, given the right input, will be called by WP anyway?

It's as simple as casing the array with `(object) $postarr`.

(*) I had to jump to the ticket to understand the reason of this change: commit messages are the right place for explanations.

strategio: @andreasciamanna, I feel that wrapping the array with `wp_slash` is more explicit. If we just cast the array to an object, we don't really have a clue why we are doing this. The reason is hidden inside `wp_update_post`.

andreasciamanna: I guess this is subjective. Looking at the code in the future, I'll still wonder why is that call needed.

Anyway, can you at least fix the commit message?

Other than the ticket title and ID, it contains zero information.

Also, the fact you need to do the same hack multiple times, creates duplication.
I think it would be better if you create a helper function and call this one instead of `wp_uodate_post`.
This helper function can then take care of calling `wp_slash` as well.

Moreover, you can write a specific test for the helper function, while the existing tests can simply check that the helper function is called with the right arguments.

strategio: @andreasciamanna, I updated the commit message.

strategio:
@andreasciamanna, do you have an idea why the pipeline is failing => https://scrutinizer-ci.com/g/OnTheGoSystems/wpml-page-builders/inspections/96e6a478-e75c-4da6-bed9-27fd1a40f3da.

I already updated `scrutinizer.yml` taking a newer version from `page-builder-gutenberg`.

src/media/shortcodes/class-wpml-page-builders-media-shortcodes-update.php
andreasciamanna: @strategio not sure why you've marked the conversation as resolved.

What about the code duplication?

Imagine we find out that `wp_slash` is not enough anymore. We would need to find all the places where we are using it and change them, with a high risk to miss some of them.

A helper function would prevent this.

strategio: @andreasciamanna, you said "at least fix the commit message" so I thought the helper was optional. Honestly, I am divided about this. We already have 2 helpers in Core: `WPML_Create_Post_Helper` and `WPML_WP_Post` which are both doing more or less the same thing (. Sometimes, we use it, sometimes not. I have the feeling that adding one more helper is overkill. And anyway, now we need to know to use `wp_slash` before `wp_update_post`. If we have the helper, all devs will need to know that it is preferable than a direct call to `wp_update_post`.

Can you confirm your idea?

````php
function wpml_update_post_array( array $post_array, $wp_error ) {
    return wp_update_post( wp_slash( $post_array, $wp_error ) );
}
```

strategio:
>In other words, compared to master, there is less code being tested, that is, the added/modified code, is not covered.

@andreasciamanna, all my changes are covered with tests.

src/media/shortcodes/class-wpml-page-builders-media-shortcodes-update.php
andreasciamanna: The ”at least” was for the solution found, since I'm not thrilled about it :)

As for the helper function, I've explained the reasons why it's needed. Code duplication is a dangerous risk I wouldn't personally take.
If you think it's not a valid reason, feel free to dismiss my request.

andreasciamanna:
> > In other words, compared to master, there is less code being tested, that is, the added/modified code, is not covered.
> 
> @andreasciamanna, all my changes are covered with tests. Maybe this happens because we just changed the `scrutinizer.yml` file and the previous code coverage reference is missing.

I don't know. When I added this rule, I didn't remember having the same issue.
Anyway, as long as the failure is about the metrics and since you've confirmed to have covered your changes with tests, the PR can be still merged.

src/media/shortcodes/class-wpml-page-builders-media-shortcodes-update.php
strategio: I did some changes as discussed @andreasciamanna, please check again.

PR #71, opened 2019-09-01
Fix error when json data is not valid

brucepearson:
Fix error when json data is not valid - https://onthegosystems.myjetbrains.com/youtrack/issue/wpmltm-3405


src/st/compatibility/class-wpml-page-builders-register-strings.php
```diff
@@ -53,10 +53,13 @@ public function register_strings( WP_Post $post, array $package ) {
...
-				$this->data_settings->convert_data_to_array( $data ),
-				$package
-			);
+			$converted = $this->data_settings->convert_data_to_array( $data );
+			if ( is_array( $converted ) ) {
+				$this->register_strings_for_modules(
+					$converted,
+					$package
+				);
+			}
```
brucepearson: Note: This code is covered by tests in the Elementor class that extends this class.

PR #73, opened 2019-09-04
Handle link type fields for gutenberg blocks

brucepearson:
Handle link type fields for gutenberg blocks - https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-6471

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -20,6 +20,12 @@
             "tests/phpunit/util/"
         ]
     },
+    "repositories": {
+        "collect": {
+            "type": "vcs",
+            "url": "https://github.com/OnTheGoSystems/collect.git"
+        }
+    },
     "require-dev": {
         "roave/security-advisories": "dev-master",
 
@@ -37,6 +43,7 @@
         "wp-coding-standards/wpcs": "^0",
 
         "m4tthumphrey/php-gitlab-api": "^9.0.0",
-        "php-http/guzzle6-adapter": "^1.1"
+        "php-http/guzzle6-adapter": "^1.1",
+        "wpml/collect": "dev-wpml-collect-rename"
     }
 }
--- a/src/st/TranslateLinks.php
+++ b/src/st/TranslateLinks.php
@@ -1,0 +1,33 @@
+<?php
+
+namespace WPML\PB;
+
+class TranslateLinks {
+
+	/**
+	 * @param \WPML_ST_String_Factory $stringFactory
+	 * @param array $activeLanguages
+	 *
+	 * @return \Closure
+	 */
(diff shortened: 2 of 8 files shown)


composer.json
```diff
@@ -20,6 +20,12 @@
             "tests/phpunit/util/"
         ]
     },
+    "repositories": [
+        {
+            "type": "vcs",
+            "url": "https://github.com/OnTheGoSystems/collect.git"
+        }
+    ],
```
andreasciamanna: @BruceOnTheGo please use the object form, as it makes easier to identify repositories and also allows manipulating them via CLI (which is not possible with the array form).

```json
"repositories": {
  "collect": {
    "type": "vcs",
    "url": "https://github.com/OnTheGoSystems/collect.git"
  }
}
```

The name of the repository can be anything you want, but "collect" (or "OnTheGoSystems/collect") is what makes more sense.

src/st/TranslateLinks.php
```diff
@@ -0,0 +1,33 @@
...
+
+class TranslateLinks {
+
+	/**
+	 * @param \WPML_ST_String_Factory $stringFactory
+	 * @param array $activeLanguages
+	 *
+	 * @return \Closure
+	 */
+	public static function getLinkTranslation( \WPML_ST_String_Factory $stringFactory, $activeLanguages ) {
```
andreasciamanna: @BruceOnTheGo I find this approach a bit hard to understand. Not really because of the code, but because of the function name.

By using a function called `getLinkTranslation`, I would expect to pass a link (as a string) as the input and get a translation as the output.

Instead, I must pass an instance of `WPML_ST_String_Factory` together with an array of active languages and then I get back another function.

Then I must call this new function with a string ID (not a link, nor a string: an ID) and then I'm completely lost, because I see no more return points.

brucepearson: @andreasciamanna How about `getLinkTranslator` because it returns a function to translate links in strings.

brucepearson: Or `getTranslatorForString`

`TranslateLinks::getTranslatorForString()`

andreasciamanna: @BruceOnTheGo the "-or" suffix may help.

composer.json
brucepearson: @andreasciamanna Done

src/st/TranslateLinks.php
brucepearson: @andreasciamanna Done

gordonsh24:
@brucepearson +1

PR #75, opened 2019-11-07
Improve support for posts which are not handled by the page builder

strategio:
- Extend interface `IWPML_Page_Builders_Data_Settings` with public
method `is_handling_post`
- Improve `is_page_builder_page` filter
- Do not include the package strings to the translation job if the body should be translated (until now, we didn't have the case when the body should be translated and we have package strings).

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-6929

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -28,20 +28,16 @@
     },
     "require-dev": {
         "roave/security-advisories": "dev-master",
-
         "10up/wp_mock": "~0.2.0",
         "lucatume/function-mocker": "1.3.4",
         "symfony/dom-crawler": "^3.1",
         "symfony/css-selector": "^3.1",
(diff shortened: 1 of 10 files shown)


strategio:
@OnTheGoSystems/wpml

dgwatkins:
+1 @strategio

strategio:
@dgwatkins, @kagg-design,

I added one more commit https://github.com/OnTheGoSystems/wpml-page-builders/pull/75/commits/94e6c6358b668bc6b5c257cb68140857419c225a => It will remove old strings from the page builder once we switch back to the defaut editor (and we won't have unneded fields in the translation job).

Please give it a look!

PR #76, opened 2019-11-22
Introduce the public filter hook "wpml_pb_post_use_translation_editor"

strategio:
It will return true if a post ID was edited with a translation editor,
or false otherwise.

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-6964

### DIFF
--- a/src/PublicApi/Hooks.php
+++ b/src/PublicApi/Hooks.php
@@ -1,0 +1,36 @@
+<?php
+
+namespace WPML\PB\PublicApi;
+
+use WPML_PB_Last_Translation_Edit_Mode;
+
+class Hooks {
+
+	/** @var WPML_PB_Last_Translation_Edit_Mode $lastEditMode */
+	private $lastEditMode;
+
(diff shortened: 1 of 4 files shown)


strategio:

src/PublicApi/Hooks.php
```diff
@@ -0,0 +1,36 @@
...
+		add_filter( 'wpml_pb_post_use_translation_editor', [ $this, 'postUseTranslationEditor' ], 10, 2 );
+	}
+
+	/**
+	 * @param bool $incomingValue
+	 * @param int  $postId
+	 *
+	 * @return bool
+	 */
+	public function postUseTranslationEditor( $incomingValue, $postId ) {
```
andreasciamanna: @strategio I think we usually called the first parameter something like `$ignored`. Not sure exactly, but I think you can find a few examples in our code base.
”incomingValue” is not clear enough, in my opinion.

andreasciamanna:
@OnTheGoSystems/wpml

PR #77, opened 2019-11-26
Load media updaters also in JSON requests.

dgwatkins:
This is needed for media updates to work when translating with ATE.

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlmedia-632


dgwatkins:
@OnTheGoSystems/wpml Please review

dgwatkins:
@OnTheGoSystems/wpml Please review

dgwatkins:
@brucepearson @strategio 
I did a small change, can you re-review?

The initial change stopped working after the latest changes in the way translation jobs are downloaded from ATE.

I paid attention at the difference of using is_admin() or not and the overhead is just attaching 2 filters that will only be used when needed so we should be ok

PR #80, opened 2020-03-03
Do not apply `wp_slash` on non-encoded data

strategio:
We started to apply `wp_slash` to fix an issue when copying some fields
on Elementor and this had a side effect on Beaver Builder (it actually
caused the same symptom than the one we fixed in Elementor).

=> Here's the related commit https://github.com/OnTheGoSystems/wpml-page-builders/commit/eb2be54fee03e982bc2727fe5a39aa7618e66186.

I think we fixed a very specific case in Elementor and possibly
something that was fixed on their side later because I could not
reproduce it after removing `return wp_slash( $data );`.

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-7141

### DIFF
--- a/tests/phpunit/tests/tm/test-wpml-pb-handle-custom-fields.php
+++ b/tests/phpunit/tests/tm/test-wpml-pb-handle-custom-fields.php
@@ -253,6 +253,19 @@ class Test_WPML_PB_Handle_Custom_Fields extends \OTGS\PHPUnit\Tools\TestCase {
 		$not_json = 'do not escape these - \\ " \'';
 		$this->assertEquals( $not_json, WPML_PB_Handle_Custom_Fields::slash_json( $not_json ) );
 	}
+
+	/**
+	 * @test
+	 * @group wpmlcore-7141
+	 */
+	public function it_does_NOT_slash_if_original_data_is_NOT_a_string() {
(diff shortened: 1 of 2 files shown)


tests/phpunit/tests/tm/test-wpml-pb-handle-custom-fields.php
```diff
@@ -253,6 +253,19 @@ public function it_does_not_slash_non_json() {
 		$not_json = 'do not escape these - \\ " \'';
 		$this->assertEquals( $not_json, WPML_PB_Handle_Custom_Fields::slash_json( $not_json ) );
 	}
+
+	/**
+	 * @test
+	 * @group wpmlcore-7141
+	 */
+	public function it_does_not_slash_is_original_data_is_not_a_string() {
```
andreasciamanna:
```suggestion
	public function it_does_NOT_slash_if_original_data_is_NOT_a_string() {
```

@strategio:

- Typo: "is" -> "if"
- Capitalized negation to make them more prominent when looking at the test names

strategio:
All done @andreasciamanna!

strategio:
@andreasciamanna, I changed the target branch and now I need you to re-approve the PR, thanks!

PR #82, opened 2020-03-27
Apply page builder string translations if those are not part of the translation job

strategio:
Some PB strings like "LINK" type are not included in the translation job
because it's automatically converted.

After adding string translations from the translation job, if we detect
that no string were added and the post is built with a page builder, we
will add the attached packages to the update list.

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-7188

### DIFF
--- a/src/st/class-wpml-pb-integration.php
+++ b/src/st/class-wpml-pb-integration.php
@@ -81,13 +81,12 @@ class WPML_PB_Integration {
 			);
 		}
 
-		foreach ( $this->strategies as $strategy ) {
-
+		$this->with_strategies( function( IWPML_PB_Strategy $strategy ) use ( $updated_packages, $post_element ) {
 			foreach ( $updated_packages as $package ) {
 				$this->factory->get_string_translations( $strategy )
 					->add_package_to_update_list( $package, $post_element->get_language_code() );
 			}
-		}
+		} );
 
 		$this->new_translations_recieved = true;
 		$this->queue_save_post_actions( $post_element->get_id(), $post_element->get_wp_object() );
@@ -129,9 +128,9 @@ class WPML_PB_Integration {
 	public function register_all_strings_for_translation( $post ) {
 		if ( $this->is_post_status_ok( $post ) && $this->is_original_post( $post ) ) {
 			$this->is_registering_string = true;
-			foreach ( $this->strategies as $strategy ) {
+			$this->with_strategies( function( IWPML_PB_Strategy $strategy ) use ( $post ) {
 				$strategy->register_strings( $post );
-			}
+			} );
 			$this->is_registering_string = false;
 		}
 	}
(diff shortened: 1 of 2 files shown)


src/st/class-wpml-pb-integration.php
```diff
@@ -195,18 +195,60 @@ public function do_shutdown_action() {
...
 	}
 
+	/**
+	 * @param callable $callable
+	 * 
+	 * @return mixed
+	 */
+	private function with_strategies( callable $callable ) {
+		foreach ( $this->strategies as $strategy ) {
+			return $callable( $strategy );
```
brucepearson: @strategio This doesn't look right. This code will only ever call the `$callable` once with the first strategy.

src/st/class-wpml-pb-integration.php
```diff
@@ -195,18 +195,60 @@ public function do_shutdown_action() {
...
+				return function( WPML_Package $package ) use ( $strategy, $targetLang ) {
+					$this->factory
+						->get_string_translations( $strategy )
+						->add_package_to_update_list( $package, $targetLang );
+				};
+			} );
+
+			wpml_collect( apply_filters( 'wpml_st_get_post_string_packages', [], $original_doc_id ) )
+				->each( $addPackageToUpdateList )
+				->first( function() {
```
brucepearson: @strategio I think you're mis-using the `first` function here. 

Maybe you could do:

```php
if ( wpml_collect( apply_filters( 'wpml_st_get_post_string_packages', [], $original_doc_id ) )
				->each( $addPackageToUpdateList )
				->isNotEmpty() ) {
    $this->new_translations_recieved = true;
}
```

strategio: Thanks @brucepearson, even better:

```php
$this->new_translations_recieved = wpml_collect( apply_filters( 'wpml_st_get_post_string_packages', [], $original_doc_id ) )
				->each( $addPackageToUpdateList )
				->isNotEmpty();
```

src/st/class-wpml-pb-integration.php
strategio: How could I miss it @brucepearson! Thanks for catching it :)

I also strenghen the test with 3 added strategies, and updated the code with inversion of functions nesting (`with_strategies` is now inside the mapping function).

src/st/class-wpml-pb-integration.php
```diff
@@ -195,18 +195,58 @@ public function do_shutdown_action() {
...
 		}
 	}
 
+	/**
+	 * @param callable $callable
+	 * 
+	 * @return mixed
+	 */
+	private function with_strategies( callable $callable ) {
+		foreach ( $this->strategies as $strategy ) {
```
brucepearson: @strategio If you make `$this->stratagies` a `wpml_collect` you can then do 
```php
$this->stratagies->each( $callable );
```

PR #85, opened 2020-04-09
Wrong processing of media shortcodes

kagg-design:
https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-7200

### DIFF
--- a/src/media/shortcodes/class-wpml-page-builders-media-shortcodes-update.php
+++ b/src/media/shortcodes/class-wpml-page-builders-media-shortcodes-update.php
@@ -50,6 +50,18 @@ class WPML_Page_Builders_Media_Shortcodes_Update implements IWPML_PB_Media_Updat
 			'post_content' => $post->post_content,
 			'tags_input'   => $tag_ids,
 		);
+
+		// Action 'wpml_tm_save_post' changes status of post. We should not touch it when duplicating.
+		$duplication = 'wpml_duplicate_dashboard' === filter_input( INPUT_POST, 'action', FILTER_SANITIZE_STRING );
+
+		if ( $duplication ) {
+			remove_action( 'wpml_tm_save_post', 'wpml_tm_save_post', 10, 3 );
+		}
+
 		wpml_update_escaped_post( $postarr );
+
+		if ( $duplication ) {
+			add_action( 'wpml_tm_save_post', 'wpml_tm_save_post', 10, 3 );
+		}
 	}
 }
--- a/tests/phpunit/bootstrap.php
+++ b/tests/phpunit/bootstrap.php
@@ -43,10 +43,11 @@ require_once __DIR__ . '/../../vendor/autoload.php';
 
 FunctionMocker::init(
 	[
-		'blacklist' => [],
-		'whitelist' => [],
+		'blacklist'             => [],
+		'whitelist'             => [],
 		'redefinable-internals' => [
 			'defined',
+			'filter_input',
 		],
 	]
 );
--- a/tests/phpunit/tests/media/shortcodes/test-wpml-page-builders-media-shortcodes-update.php
+++ b/tests/phpunit/tests/media/shortcodes/test-wpml-page-builders-media-shortcodes-update.php
@@ -1,5 +1,7 @@
 <?php
 
+use tad\FunctionMocker\FunctionMocker;
+
 /**
  * @group media
  */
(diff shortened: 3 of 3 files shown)


kagg-design:
@OnTheGoSystems/wpml Please review.

strategio:
@kagg-design, I have a doubt about your fix. After the inner `foreach`, there's some code to set `$shortcode_data[]`.

I think it would be used in that case:

`[my_shortcode images="123,456"]Some content[/my_shortcode]`

With the following configuration:

```php
<shortcode>
            <tag>my_shortcode</tag>
            <attributes>
                <attribute type="media-ids">images</attribute>
            </attributes>
</shortcode>
```

With your fix, we will miss the string "Some content" and won't be able to translate it.

It's hard to understand the root cause (I even checked the YT ticket), could you please share more details?

kagg-design:
@strategio Let me first explain the nature of the bug with `continue 2`.

Here is the breakpoint in `wpml_config_filter()`.



We can see that fusion builder has 147 shortcodes. OK.

The goal of `wpml_config_filter()` is to separate shortcodes for non-media/media and write them to relevant options.

What we have in master in the `update_shortcodes_config()` - 146 shortcodes are considered as non/media:



What we have in master in the `update_medai_shortcodes_config()` - 95 shortcodes are considered as media:



This is the bug by itself. Only 5 shortcodes in fusion builder are media shortcodes.

Now, I use `continue 2` and have the following results:

`update_shortcodes_config()` - 125 shortcodes are considered as non/media.



`update_media_shortcodes_config()` - 5 shortcodes considered as media (and this is exact number).



This bug happened because in tests, line 44 and 98 in `vendor/wpml/page-builders/src/st/strategy/shortcode/class-wpml-pb-config-import-shortcode.php` were not covered by tests in master, because `is_media_attribute()` always returned `true` in tests, because there were no proper test data created.

kagg-design:
@pierre Now, what happens when you duplicate a post in TM dashboard. Execution comes here:



and `has_media_shortcode()` returns `true` for the shortcode `fusion_builder_column`. This is wrong. It is not a media shortcode.

Basing on the result of `has_media_shortcode()`, further code marks post with not "duplicated" icon, but with "green checkmark" icon, and it is the root issue, reported in the ticket.

I don't know what is the logic in our code to mark posts with media shortcodes by a green checkmark, it is the subject of a separate investigation. I think something is behind of it.

Anyway, answer that post without any media shortcodes has media shortcodes, is wrong, and has to be fixed.

kagg-design:
@pierre I am working on your case in the comment `[my_shortcode images="123,456"]Some content[/my_shortcode]` and will add a fix a bit later.

strategio:
> @strategio Let me first explain the nature of the bug with `continue 2`.
> ...
> This bug happened because in tests, line 44 and 98 in `vendor/wpml/page-builders/src/st/strategy/shortcode/class-wpml-pb-config-import-shortcode.php` were not covered by tests in master, because `is_media_attribute()` always returned `true` in tests, because there were no proper test data created.

@kagg-design, you might think that a shortcode can be either a media or a non-media one but I think this is not always true. The shortcode configuration can be used both (partially) for a non-media and media options.

I'll give you a theorical example: Let's say that the shortcode [fusion_popover](https://github.com/OnTheGoSystems/wpml-config/blob/master/avada/wpml-config.xml#L423) contains an image inside the content:

```
[fusion_popover attrs...]Some content <img src="http://path/to/image">[/fusion_popover]
```

`fusion_popover` will be handle in the media process because it could contain an image URL that we would like to convert to the translated one.

That's why we introduced the config attribute `ignore-content`.

This was implemented some time ago now and I confess it's hard for me to remember exactly the whole thing, but you should take into consideration my example and maybe test it. Perhaps you should also run the CC tests with your fix.

kagg-design:
@strategio After further investigation of the code, I think that purpose of those two methods is different. `update_shortcodes_config` should strip media attributes from shortcodes.  `update_media_shortcodes_config` should collect only those shortcodes which have at least one media attribute.

Very sad that we don't have even a single line in the description of the methods. So now, we just guess what was the initial purpose of those methods and what should be the expected result.

I have made fixes according to the above mentioned understanding, and added tests.

> [my_shortcode images="123,456"]Some content[/my_shortcode] is tested manually and works properly now.

kagg-design:
@pierre I have also noticed that as input, we have duplicated shortcodes, and I don't know the reason. I have just added `array_unique_multi()` to remove duplicates.

kagg-design:
@strategio `[fusion_popover attrs...]Some content [/fusion_popover]` does not have `ignore-content="1"` in tag description, so I tested with `[fusion_gallery image_ids="100,200"]Some content [/fusion_gallery]`. Everything is fine, it does not appear in CTE.

strategio:
@kagg-design, looking at your last change, I still think we might break something else. Let's consider the following shortcode:

```html
[my_shortcode]http://path/to/my/image[/my_shortcode]
```

With the following config:

```xml
<shortcode>
    <tag type="media-url">my_shortcode</tag>
</shortcode>
```

I looked at our `res/xsd/wpml-config.xsd` and this configuration is allowed so I believe we did it for a good reason:

```xsd
<xs:complexType name="wpml-shortcode-tag">
    <xs:simpleContent>
        <xs:extension base="xs:string">
            <!-- todo: define restriction with enumeration -->
            <xs:attribute type="xs:string" name="encoding" use="optional"/>
            <xs:attribute type="xs:string" name="encoding-condition" use="optional"/>
            <xs:attribute type="xs:string" name="raw-html" use="optional"/>
            <xs:attribute name="ignore-content" use="optional">
                <xs:simpleType>
                    <xs:restriction base="xs:integer">
                        <xs:enumeration value="0"/>
                        <xs:enumeration value="1"/>
                    </xs:restriction>
                </xs:simpleType>
            </xs:attribute>
            <xs:attribute name="type" use="optional">
                <xs:simpleType>
                    <xs:restriction base="xs:string">
                        <xs:enumeration value="media-url"/>
                    </xs:restriction>
                </xs:simpleType>
            </xs:attribute>
        </xs:extension>
    </xs:simpleContent>
</xs:complexType>
```

strategio:
@kagg-design, now I don't know if it's a bug in our code or a wrong definition in our config file. As you already noticed duplicated shortcode definition, it means that our config files might not be very clean.

I believe that we should also prevent any issue if the config is not correct, but I am not sure exactly where to fix it, maybe at the place where you actually saw the fatal error.

kagg-design:
@strategio I have finished this crunchy debugging, and have found the solution. Please review again.

kagg-design:
After discussion, I close this PR in favour of https://git.onthegosystems.com/wpml/wpml-translation-management/-/merge_requests/1599

PR #87, opened 2020-05-09
Allow translating content as page builder with mixed shortcodes and regular text

strategio:
This happens when we have a content like below:

`[my_shortcode title="This is a title"] Hello there`

If we have at least one declared shortcode in `wpml-config.xml` (i.e.
"my_shortcode" in this situation), we will wrap the whole content to
translate it as a page builder. As a result in the translation job,
we'll have 2 strings: "This is a title" and "Hello there".

However, if `my_shortcode` is not a declared shortcode, we will not wrap the content and no page builder string will extracted. As a consequence, we will still display the fill post body in the translation job.

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-6971

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -20,12 +20,6 @@
             "tests/phpunit/util/"
         ]
     },
-    "repositories": {
-        "collect": {
-            "type": "vcs",
-            "url": "https://github.com/OnTheGoSystems/collect.git"
-        }
-    },
     "require-dev": {
         "roave/security-advisories": "dev-master",
         "10up/wp_mock": "~0.2.0",
@@ -40,6 +34,8 @@
         "wp-coding-standards/wpcs": "^0",
         "m4tthumphrey/php-gitlab-api": "^9.0.0",
         "php-http/guzzle6-adapter": "^1.1",
-        "wpml/collect": "dev-wpml-collect-rename"
+        "wpml/collect": "dev-wpml-collect-rename",
+        "wpml/fp": "dev-master",
+        "wpml/wp": "dev-master"
     }
 }
--- a/src/st/strategy/shortcode/class-wpml-pb-shortcode-content-wrapper.php
+++ b/src/st/strategy/shortcode/class-wpml-pb-shortcode-content-wrapper.php
@@ -1,8 +1,14 @@
 <?php
 
+use WPML\FP\Logic;
+use WPML\FP\Maybe;
+use WPML\FP\Str;
+use function WPML\FP\pipe;
+
 class WPML_PB_Shortcode_Content_Wrapper {
 
-	const WRAPPER_SHORTCODE_NAME = 'wpml_string_wrapper';
+	const WRAPPER_SHORTCODE_NAME  = 'wpml_string_wrapper';
+	const GUTENBERG_OPENING_START = '<!-- wp:';
 
 	/** @var string $content */
 	private $content;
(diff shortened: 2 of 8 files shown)


tests/phpunit/tests/st/strategy/shortcode/test-wpml-pb-register-shortcodes.php
```diff
@@ -41,9 +42,6 @@ public function register_shortcode_strings( $encode ) {
 			'args'  => array( $post_id, WPML_PB_Integration::MIGRATION_DONE_POST_META, true ),
 		) );
 
-
-		$dummy_content = rand_long_str( 100 ) . $shortcode . rand_long_str( 100 );
```
strategio, line 45: I removed the content before and after, because now it would be wrapped and we would register 2 more strings. This case is considered in the next test: `it_should_register_strings_in_non_wrapped_content`.

strategio:
FTR, I reused the content wrapper logic that I created in https://github.com/OnTheGoSystems/wpml-page-builders/pull/48.

src/st/strategy/shortcode/class-wpml-pb-shortcode-content-wrapper.php
```diff
@@ -250,4 +250,65 @@ private function insert_wrappers() {
...
 	}
+
+	/**
+	 * @param string $content
+	 * @param array  $shortcodes
+	 *
+	 * @return string
+	 */
+	public static function maybeWrap( $content, array $shortcodes ) {
+		return self::shouldBeWrapped( $content, $shortcodes ) ? self::wrap( $content ) : $content;
```
brucepearson, line 261: @strategio You could also inline the functions and then use a maybe pipeline.

```php

$containsOneShortcode = pipe( Str::match( '/' . get_shortcode_regex( $shortcodes ) . '/s' ), Fns::isEmpty(), Fns::not() ); 

$withoutShortcodesIsDifferent = curryN( 2, function( $shortcodes, $content ) {
    ....
    return (bool) ....;
} ) );

return Maybe::of( $content )
   ->filter( $containsOneShortcode( $shortcodes) )
   ->filter( $withoutShortcodesIsDifferent )
   ->map( [ self::class 'wrap' ] )
   ->getOrElse( $content );
```

strategio: Thanks for the tip @brucepearson :)

I implemented this but now I need a way to download the composer dependencies coming from GitLab because the FP repo is private. I will do this as soon as Andrea gives me a token.

strategio:
I added one more commit (https://github.com/OnTheGoSystems/wpml-page-builders/pull/87/commits/1589fba57bf26b51cc854394000c6c977404fd36) to not register the Gutenberg block comments.

Currently we don't support both GB blocks and shortcodes (one PB will overwrite the other one) but this is already an issue. For now, we will just not register the comments as strings.

strategio:
Done @brucepearson (I included the FP and WP packages). Please follow your review.

composer.json
```diff
@@ -40,6 +48,8 @@
         "wp-coding-standards/wpcs": "^0",
         "m4tthumphrey/php-gitlab-api": "^9.0.0",
         "php-http/guzzle6-adapter": "^1.1",
-        "wpml/collect": "dev-wpml-collect-rename"
+        "wpml/collect": "dev-wpml-collect-rename",
+        "wpml/fp": "dev-develop",
```
brucepearson, line 52: @strategio Please update fp and wp to master branches.

src/st/strategy/shortcode/class-wpml-pb-shortcode-content-wrapper.php
```diff
@@ -250,4 +255,62 @@ private function insert_wrappers() {
...
 	}
+
+	/**
+	 * @param string $content
+	 * @param array  $shortcodes
+	 *
+	 * @return string
+	 */
+	public static function maybeWrap( $content, array $shortcodes ) {
+		/** @see WPML_Gutenberg_Integration::GUTENBERG_OPENING_START */
```
brucepearson, line 266: @strategio Why have this comment when you can code it:
```php
$notGutenbergContent  = pipe( Str::includes( WPML_Gutenberg_Integration::GUTENBERG_OPENING_START ), Logic::not() );
```

strategio: @brucepearson, the class `WPML_Gutenberg_Integration` is in another project, that's why I opted for a PHPDoc. Otherwise I would have to create a stub which is overkill just for this.

brucepearson: @strategio So the comment confused me more than helping. Maybe it's better to just add a duplicate class constant in this class.

composer.json
```diff
@@ -24,6 +24,14 @@
...
             "type": "vcs",
             "url": "https://github.com/OnTheGoSystems/collect.git"
+        },
+        "wp" : {
+            "type": "vcs",
+            "url": "https://git.onthegosystems.com/wpml-packages/wp.git"
+        },
+        "fp" : {
+            "type": "vcs",
+            "url": "https://git.onthegosystems.com/wpml-packages/fp.git"
```
brucepearson, line 34: @strategio You shouldn't need these now as they should come from packagist.

PR #88, opened 2020-05-12
Prevent running "pro_translation_completed" action when not a post job

strategio:
https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-7246

### DIFF
--- a/src/tm/class-wpml-tm-page-builders.php
+++ b/src/tm/class-wpml-tm-page-builders.php
@@ -92,11 +92,15 @@ class WPML_TM_Page_Builders {
 	}
 
 	/**
-	 * @param int      $new_post_id
+	 * @param int      $new_element_id
 	 * @param array    $fields
 	 * @param stdClass $job
 	 */
-	public function pro_translation_completed_action( $new_post_id, array $fields, stdClass $job ) {
+	public function pro_translation_completed_action( $new_element_id, array $fields, stdClass $job ) {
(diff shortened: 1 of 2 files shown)


src/tm/class-wpml-tm-page-builders.php
```diff
@@ -92,11 +92,15 @@ public function translation_job_data_filter( array $translation_package, $post )
...
 
 	/**
-	 * @param int      $new_post_id
+	 * @param int      $new_element_id
 	 * @param array    $fields
 	 * @param stdClass $job
 	 */
-	public function pro_translation_completed_action( $new_post_id, array $fields, stdClass $job ) {
+	public function pro_translation_completed_action( $new_element_id, array $fields, stdClass $job ) {
+		if ( 'post' !== $job->element_type_prefix ) {
```
brucepearson, line 100: @strategio I assume that this change now means that this code only handles strings that are part of the post body like shortcodes and Gutenberg blocks.

strategio: Yes I confirm @brucepearson. Currently we have 3 `element_type_prefix`: `post`, `package`, `string`.

This code should run only for the `post` type which will come with "package string fields" in the job.

strategio: That's also why I changed the variable name `$new_post_id` to `$new_element_id` because it was confusing.

PR #89, opened 2020-05-22
Auto update translations when the original is updated no string changes

strategio:
Goes with https://git.onthegosystems.com/wpml/wpml-translation-management/-/merge_requests/1615.

- The content hash used for the translation status will be made of a
concatenation of the package string hashes.
- When updating the original with no string changes, we will resave the
completed translations (applying the string translations on the updated
content).

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-7263

### DIFF
--- a/src/AutoUpdate/Hooks.php
+++ b/src/AutoUpdate/Hooks.php
@@ -1,0 +1,94 @@
+<?php
+
+namespace WPML\PB\AutoUpdate;
+
+use WPML\FP\Fns;
+use WPML\FP\Logic;
+use WPML\FP\Lst;
+use WPML\FP\Maybe;
+use WPML\FP\Obj;
+use WPML\FP\Relation;
+use function WPML\FP\invoke;
+use function WPML\FP\pipe;
+
+class Hooks implements \IWPML_Backend_Action, \IWPML_Frontend_Action, \IWPML_DIC_Action {
+
+	const HASH_SEP = '-';
+
+	/** @var \WPML_PB_Integration $pbIntegration */
+	private $pbIntegration;
+
+	/** @var \WPML_Translation_Element_Factory $elementFactory */
+	private $elementFactory;
+
+	public function __construct(
+		\WPML_PB_Integration $pbIntegration,
+		\WPML_Translation_Element_Factory $elementFactory
+	) {
+		$this->pbIntegration     = $pbIntegration;
+		$this->elementFactory    = $elementFactory;
+	}
+
+	public function add_hooks() {
+		add_filter( 'wpml_tm_post_md5_content', [ $this, 'getMd5ContentFromPackageStrings' ], 10, 2 );
+		add_action( 'wpml_after_save_post', [ $this, 'resaveTranslationsAfterSavePost' ], 10, 4 );
+	}
+
+	/**
+	 * @param string   $content
+	 * @param \WP_Post $post
+	 *
+	 * @return string
+	 */
+	public function getMd5ContentFromPackageStrings( $content, $post ) {
+		// $joinPackageStringHashes :: \WPML_Package → string
+		$joinPackageStringHashes = pipe(
+			Obj::propOr( [], 'string_data' ),
+			Obj::keys(),
+			Lst::sort( Relation::gt() ),
+			Lst::join( self::HASH_SEP )
+		);
+
+		return Maybe::of( $post->ID )
+			->map( [ self::class, 'getPackages' ] )
+			->map( Fns::map( $joinPackageStringHashes ) )
+			->filter()
+			->map( Lst::join( self::HASH_SEP ) )
+			->getOrElse( $content );
+	}
+
+	/**
+	 * @param int $postId
+	 *
+	 * @return \WPML_Package[]
+	 */
+	public static function getPackages( $postId ) {
+		return apply_filters( 'wpml_st_get_post_string_packages', [], $postId );
+	}
+
+	/**
+	 * @param int         $postId
+	 * @param int         $trid
+	 * @param string      $lang
+	 * @param string|null $sourceLang
+	 */
+	public function resaveTranslationsAfterSavePost( $postId, $trid, $lang, $sourceLang ) {
+		if ( $sourceLang || ! self::getPackages( $postId ) ) {
+			return;
+		}
+
+		// $ifOriginal :: \WPML_Post_Element → bool
+		$ifOriginal = pipe( invoke( 'get_source_language_code' ), Logic::not() );
+
+		// $ifCompleted :: \WPML_Post_Element → bool
+		$ifCompleted = pipe( [ TranslationStatus::class, 'get' ], Relation::equals( ICL_TM_COMPLETE ) );
+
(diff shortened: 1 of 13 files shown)


src/Container/Config.php
```diff
@@ -0,0 +1,13 @@
+<?php
+
+namespace WPML\PB\Container;
+
+class Config {
+
+	public static function getSharedClasses() {
+		return [
+			'\WPML_PB_Factory',
+			'\WPML_PB_Integration',
```
strategio, line 10: This `\WPML_PB_Integration` class has a lot of code in it and it's getting hard to extend and test properly. Having a unique instance of this class allows to use is as a dependency in `WPML\PB\AutoUpdate\Hooks`.

src/st/class-wpml-pb-factory.php
```diff
@@ -4,11 +4,13 @@
...
 class WPML_PB_Factory {
 
+	/** @var wpdb */
 	private $wpdb;
+	/** @var SitePress */
 	private $sitepress;
 	private $string_translations = array();
 
-	public function __construct( $wpdb, $sitepress ) {
+	public function __construct( wpdb $wpdb, SitePress $sitepress ) {
```
strategio, line 13: Now that this class is instantiated with the DIC, we need to have type hints.

src/st/class-wpml-pb-loader.php
```diff
@@ -64,12 +68,24 @@ public function __construct(
...
 	}
 
-	private static function load_hooks() {
+	/**
+	 * @param bool $has_page_builder_strategy
+	 */
+	private static function load_hooks( $has_page_builder_strategy ) {
 		$hooks = [
-			WPML_PB_Handle_Post_Body::class,
 			WPML\PB\Compatibility\Toolset\Layouts\HooksFactory::class,
```
strategio, line 76: This class for Toolset will be remove in another task (it has been migrated there now).

src/AutoUpdate/Hooks.php
```diff
@@ -0,0 +1,95 @@
...
+	 */
+	public function getMd5ContentFromPackageStrings( $content, $post ) {
+		if ( ! self::getPackages( $post->ID ) ) {
+			return $content;
+		}
+
+		// $joinPackageStringHashes :: \WPML_Package → string
+		$joinPackageStringHashes = pipe(
+			Obj::propOr( [], 'string_data' ),
+			'array_keys',
```
gordonsh24, line 50: @strategio I also defined recently: `Obj::keys`. You can use it if you wish.

src/AutoUpdate/Hooks.php
```diff
@@ -0,0 +1,95 @@
...
+			Lst::sort( Relation::gt() ),
+			Lst::join( self::HASH_SEP )
+		);
+
+		return wpml_collect( self::getPackages( $post->ID ) )
+			->map( $joinPackageStringHashes )
+			->filter()
+			->implode( self::HASH_SEP );
+	}
```
brucepearson, line 59: @strategio You could try this. `getPackages` will need to be public which I don't think is a problem because it's just a helper function and has no state.

The `->map( map( ... ) )` is needed because the Maybe holds an array at this point

```php
	public function getMd5ContentFromPackageStrings( $content, $post ) {

		// $joinPackageStringHashes :: \WPML_Package → string
		$joinPackageStringHashes = pipe(
			Obj::propOr( [], 'string_data' ),
			'array_keys',
			Lst::sort( Relation::gt() ),
			Lst::join( self::HASH_SEP )
		);

		return Maybe::of( $post->ID )
			->map( [ self::class, 'getPackages' ] ) 
			->map( map( $joinPackageStringHashes ) )
			->filter()
			->map( Lst::join( self::HASH_SEP ) )
			->getOrElse( $content );
	}
```

src/AutoUpdate/TranslationStatus.php
```diff
@@ -0,0 +1,22 @@
...
+	 * @param WPML_Post_Element $element
+	 *
+	 * @return int|null
+	 */
+	public static function get( WPML_Post_Element $element ) {
+		try {
+			return make( '\WPML_TM_Translation_Status' )
+				->filter_translation_status( null, $element->get_trid(), $element->get_language_code() );
+		} catch ( Exception $e ) {}
+	}
```
brucepearson, line 21: @strategio I'm not sure what will make it throw an exception but I assume it's `WPML_TM_Translation_Status` not being available.

I'm using this pattern for this:
```php
return Maybe::fromNullable( make( '\WPML_TM_Translation_Status' ) )
		->map( invoke( 'filter_translation_status' )->with( null, $element->get_trid(), $element->get_language_code() ) )
		->getOrElse( null );
```

brucepearson: @strategio Notice this class has nothing to do with auto update except for being a helper. Ideally we should move this type of thing into a library. It would have to go into the core-api library so it's available even if TM is not available.

tests/phpunit/tests/AutoUpdate/TestTranslationStatus.php
```diff
@@ -0,0 +1,59 @@
...
+/**
+ * @group auto-update
+ */
+class TestTranslationStatus extends TestCase {
+
+	/**
+	 * @test
+	 */
+	public function itGetsAndReturnNullOnException() {
+		\WP_Mock::userFunction( 'WPML\Container\make' )->andThrow( new Exception() );
```
brucepearson, line 17: @strategio I don't think `make` throws an exception in reality. It just returns null if the class doesn't exist.

src/AutoUpdate/TranslationStatus.php
strategio: Yes, that's the reason @brucepearson. We had one similar issue in the past so I decided to use this small wrapper class.

I also wanted to introduce this in the core-api library, but I was unsure as I don't have enough background with it. For instance, I decided to pass it a `\WPML_Post_Element` because it's convenient for me here, but I am not sure I would do that way in the library. I would rather try to use a convenient signature for what we already have in `WPML\Element\API\PostTranslations`.

I will try in a separate task (as a training case).

tests/phpunit/tests/AutoUpdate/TestTranslationStatus.php
strategio: Right @brucepearson, `make` will return an Exception if it can't resolve the dependencies. I adjusted my code accordingly.

src/AutoUpdate/Hooks.php
strategio: I initially tried something similar but I had some problems. I applied your suggestion.

src/AutoUpdate/Hooks.php
strategio: Right @gordonsh24, applied your suggestion!

strategio:
@brucepearson, @gordonsh24, I added a 2nd commit so I can get refreshed strings from the DB. Please check this!

brucepearson:
@strategio Looks good to me. +1

PR #91, opened 2020-06-08
Allow shortcodes to be processed along with gutenberg blocks

brucepearson:


src/st/class-wpml-pb-integration.php
```diff
@@ -272,15 +282,20 @@ public function migrate_location( $post_id ) {
...
-					$strategy->migrate_location( $post_id, $post->post_content );
-				} );
+				$this->with_strategies( invoke( 'migrate_location' )->with( $post_id, $post->post_content ) );
 			}
 
 			$this->mark_migrate_location_done( $post_id );
 		}
 	}
 
+	public function register_strings_in_content( $registered, $post_id, $content ) {
```
strategio, line 292: @brucepearson, I can see that you did not add a PHPDoc everywhere, it's missing in several places. Could you fix it please?

brucepearson: @strategio I added.

PR #92, opened 2020-06-11
Fix edge case when there's only one block on the page and it contains…

brucepearson:
… a shortcode


src/st/class-wpml-pb-string-translation-by-strategy.php
```diff
@@ -89,4 +89,20 @@ public function add_package_to_update_list( WPML_Package $package, $language ) {
...
+	 *
+	 * @return bool
+	 */
+	private function shouldHandlePackage( $package ) {
+		$strategyPackageKind = $this->strategy->get_package_kind();
+
+		return apply_filters(
+			'wpml_pb_should_handle_package',
+			$package->kind == $strategyPackageKind,
+			$package->kind,
```
andreasciamanna, line 104: @BruceOnTheGo as a second argument to the filter, I would pass the whole `$package` object.

brucepearson: I prefer to just pass the `kind` to keep the api more focused.

PR #93, opened 2020-06-23
Handle edge case where there is only one Gutenberg block with a shortcode

brucepearson:
This is the best I could do. My initial try was to use hooks so the Gutenberg module could override the shortcode logic but this did not work. So I added a simple class here to detect and handle the case for shortcodes.

### DIFF
--- a/src/st/ShortCodesInGutenbergBlocks.php
+++ b/src/st/ShortCodesInGutenbergBlocks.php
@@ -2,6 +2,8 @@
 
 namespace WPML\PB;
 
+use WPML\FP\Fns;
+
 /**
  * Class ShortCodesInGutenbergBlocks
  * @package WPML\PB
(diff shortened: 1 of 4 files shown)


src/st/ShortCodesInGutenbergBlocks.php
```diff
@@ -0,0 +1,53 @@
...
+		}
+
+		return $package_data;
+	}
+
+	public static function normalizePackages( array $packagesToUpdate ) {
+		if ( count( $packagesToUpdate ) > 1 ) {
+			// If we have more than one package then we don't need to 'Force' it.
+			// The normal Gutenberg package will update all translations correctly.
+			foreach ( $packagesToUpdate as $id => $package_data ) {
```
gordonsh24, line 44: @brucepearson how about `array_filter` to avoid that ugly `unset` call?

```
$packagesToUpdate  = array_filter($packagesToUpdate , function($package_data) {
   return $package_data['package']->kind !== self::FORCE_GUTEBERG;
});
```

brucepearson: @gordonsh24 Good idea.

PR #94, opened 2020-07-07
Clean up unused shortcode strings when in Gutenberg blocks

brucepearson:
Clean up unused shortcode strings when in Gutenberg blocks - https://onthegosystems.myjetbrains.com/youtrack/issue/wpmltm-3843

### DIFF
--- a/src/st/class-wpml-pb-integration.php
+++ b/src/st/class-wpml-pb-integration.php
@@ -1,7 +1,9 @@
 <?php
 
 use \WPML\FP\Fns;
+use WPML\PB\Shortcode\StringCleanUp;
 use function WPML\FP\invoke;
+use function WPML\Container\make;
 
 /**
  * Class WPML_PB_Integration
@@ -170,8 +172,10 @@ class WPML_PB_Integration {
 		add_filter( 'wpml_tm_translation_job_data', array( $this, 'rescan' ), 9, 2 );
 
 		add_action( 'wpml_pb_register_all_strings_for_translation', [ $this, 'register_all_strings_for_translation' ] );
-		add_filter( 'wpml_pb_register_strings_in_content', [ $this, 'register_strings_in_content' ], 10, 3 );
+		add_filter( 'wpml_pb_register_strings_in_content', [ $this, 'register_strings_in_content' ], 10, 4 );
 		add_filter( 'wpml_pb_update_translations_in_content', [ $this, 'update_translations_in_content'], 10, 2 );
+
+		add_filter( 'wpml_pb_get_string_clean_up', [ $this, 'get_string_clean_up' ], 10, 2 );
 	}
 
 	/**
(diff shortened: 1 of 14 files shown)


src/st/strategy/api-hooks/class-wpml-pb-api-hooks-strategy.php
```diff
@@ -19,13 +19,13 @@ public function register_strings( $post ) {
...
 	/**
-	 * @param bool $registered
 	 * @param string|int $post_id
 	 * @param string $content
+	 * @param WPML\PB\Shortcode\StringCleanUp $stringCleanUp
 	 *
 	 * @return bool
 	 */
-	public function register_strings_in_content( $post_id, $content, $do_cleanup ) {}
+	public function register_strings_in_content( $post_id, $content, WPML\PB\Shortcode\StringCleanUp $stringCleanUp ) {}
```
strategio, line 28: @brucepearson, it's a bit strange to see `WPML\PB\Shortcode\StringCleanUp` (shortcode) in the API strategy. An interface would be more suitable (`WPML\PB\StringCleanUp`). Or maybe, if you want to keep it simple, you can just keep the type in the PHPDoc and remove the type hint in the signature.

src/st/class-wpml-pb-integration.php
```diff
@@ -314,6 +319,14 @@ public function get_factory() {
 		return $this->factory;
 	}
 
+	public function get_string_clean_up( $_, $postId ) {
```
strategio, line 322: This looks like a leftover @brucepearson.

src/st/strategy/api-hooks/class-wpml-pb-api-hooks-strategy.php
strategio: Well, actually this `WPML\PB\Shortcode\StringCleanUp` is not specific to shortcodes because we are passing the strategy as a dependency. So you just need to use a better namespace ( e.g. `WPML\PB\StringCleanUp`).

src/st/class-wpml-pb-integration.php
andreasciamanna: @strategio why? It's hooked above: `add_filter( 'wpml_pb_get_string_clean_up', [ $this, 'get_string_clean_up' ], 10, 2 );`

strategio: Ah right @andreasciamanna, my bad :)

src/st/strategy/api-hooks/class-wpml-pb-api-hooks-strategy.php
brucepearson: `WPML\PB\Shortcode\StringCleanUp $stringCleanUp` is clean up code specific to shortcodes so I don't want to remove it from the namespace. The API strategy doesn't use it or implement any functionality for `register_strings_in_content`.
So I think it's ok to keep this as it is.

strategio: OK, you call @brucepearson!

PR #99, opened 2020-07-23
Force to load TM element translations hooks with `wpml_tm_load_element_translations`

strategio:
This is required because some page builders (e.g. Cornerstone) might run
on the frontend and TM is not fully loaded in that context.

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmltm-3867

### DIFF
--- a/src/AutoUpdate/TranslationStatus.php
+++ b/src/AutoUpdate/TranslationStatus.php
@@ -15,6 +15,8 @@ class TranslationStatus {
 	 * @return int|null
 	 */
 	public static function get( WPML_Post_Element $element ) {
+		wpml_tm_load_element_translations();
+
 		return Maybe::fromNullable( make( '\WPML_TM_Translation_Status' ) )
 			->map( invoke( 'filter_translation_status' )->with( null, $element->get_trid(), $element->get_language_code() ) )
 			->getOrElse( null );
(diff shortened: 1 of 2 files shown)


src/AutoUpdate/TranslationStatus.php
```diff
@@ -15,6 +15,8 @@ class TranslationStatus {
 	 * @return int|null
 	 */
 	public static function get( WPML_Post_Element $element ) {
+		wpml_tm_load_element_translations();
```
brucepearson, line 18: @strategio I don't think this is the right place for this. Adding it here makes it very hard to understand the dependencies.

Basically `$wpml_tm_element_translations` is required in `WPML_TM_Translation_Status` so I would move the initialisation to there.

```
	public function filter_translation_status( $status, $trid, $target_lang_code ) {
		/** @var WPML_TM_Element_Translations $wpml_tm_element_translations */
		$wpml_tm_element_translations = wpml_tm_load_element_translations();
```

strategio: You are completely right @brucepearson (added in https://git.onthegosystems.com/wpml/wpml-translation-management/-/merge_requests/1656/diffs?commit_id=b2851e1f7acb74554dd337febda0404b9e79242f).

strategio:
Closing here, fixed in TM the way suggested by Bruce.

PR #103, opened 2020-07-28
Let the auto-update process use its own save post queue based on `wpml_tm_save_post`

strategio:
First I reverted the commit afbad61, it was an optimization and this is the reason of the defect.

The main save post queue in the integration class is built with the `save_post` action which has a very wide scope.

For the auto-update feature we need a queue built with the
`wpml_tm_save_post` action which is more restrictive (excluding
attachments and nav items for instance). Also, it's not always fired
depending on the context and we'll catch only the needed posts.

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-7412


strategio:
@OnTheGoSystems/compatibility, @OnTheGoSystems/wpml please review.

PR #106, opened 2020-08-07
Support `label` attrib in shortcodes and gutenberg blocks

gordonsh24:


gordonsh24:

It lets us define lables for shortcodes defined in `wpml-config.xml` in this way:

```
    <shortcodes>
        <shortcode>
            <tag>vc_text_separator</tag>
            <attributes>
                <attribute label="Seperator title">title</attribute>
            </attributes>
        </shortcode>
        <shortcode>
            <tag label="Jakub's shortode">jakub_shortcode</tag>
        </shortcode>
    </shortcodes>
```

They are displayed in ATE and CTE near a field.

src/st/strategy/shortcode/class-wpml-pb-shortcode-strategy.php
```diff
@@ -88,6 +89,14 @@ public function get_shortcode_attribute_type( $tag, $attribute ) {
 		return 'LINE';
 	}
 
+	public function get_shortcode_attribute_label( $tag, $attribute ) {
+		if ( $attribute === 'content' ) {
+			return isset( $this->shortcodes[ $tag ]['label'] ) ? $this->shortcodes[ $tag ]['label'] : '';
+		} else {
+			return isset( $this->shortcodes[ $tag ]['attributes'][ $attribute ]['label'] ) ? $this->shortcodes[ $tag ]['attributes'][ $attribute ]['label'] : '';
+		}
+	}
```
strategio, line 98: @gordonsh24, for this private method (new code), I would use the FP library:

```php
$labelPath = 'content' === $attribute ? [ $tag, 'label' ] : [ $tag, 'attributes', $attribute, 'label' ];

return Obj::path( $labelPath, $this->shortcodes );
```

One day we can simplify the whole code here (but another day :D )

strategio: BTW, according to what Andrea said, it might be a good time to tag the FP and WP packages we're using here.

gordonsh24: @strategio Good idea, I used `pathOr`. I will create tags when it passes QA tests.

brucepearson (approved):

PR #108, opened 2020-09-01
Delegate only the translation statuses update from the original

strategio:
Instead of defering the whole `wpml_tm_save_post` function to the
shutdown, we will just take control of the translation status updater (a
callable).

It also simplifies the logic because we don't need to rely on the post
save queue from the integration class.

Also added a commit to re-activate the "translation auto update" feature.

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-7422

### DIFF
--- a/src/AutoUpdate/Hooks.php
+++ b/src/AutoUpdate/Hooks.php
@@ -7,6 +7,7 @@ use WPML\FP\Logic;
 use WPML\FP\Lst;
 use WPML\FP\Maybe;
 use WPML\FP\Relation;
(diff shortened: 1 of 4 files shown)


gordonsh24 (approved):
@strategio Cunning solution! +1

strategio:
@mrudzki91, I will request your review here too since you did it in https://git.onthegosystems.com/wpml/wpml-translation-management/-/merge_requests/1701 :)

mrudzki91 (approved):

mrudzki91:
@strategio everything looks nice! 👍 :)

PR #109, opened 2020-09-08
Introduce filter hook `wpml_pb_is_auto_update_activated`

strategio:
**Goes with https://github.com/OnTheGoSystems/wpml-page-builders-elementor/pull/123**

This filter will return `true` when the "translation auto-update"
feature is activated. This helps to keep an "atomic" logic inside the
`WPML\PB\AutoUpdate\Hooks` class and make it easy to deactivate if we
have any kind of trouble.

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-7544

### DIFF
--- a/src/AutoUpdate/Hooks.php
+++ b/src/AutoUpdate/Hooks.php
@@ -35,6 +35,7 @@ class Hooks implements \IWPML_Backend_Action, \IWPML_Frontend_Action, \IWPML_DIC
 
 	public function add_hooks() {
 		if ( $this->isTmLoaded() ) {
+			add_filter( 'wpml_pb_auto_update_enabled', '__return_true' );
 			add_filter( 'wpml_tm_delegate_translation_statuses_update', [ $this, 'enqueueTranslationStatusUpdate'], 10, 3 );
 			add_filter( 'wpml_tm_post_md5_content', [ $this, 'getMd5ContentFromPackageStrings' ], 10, 2 );
 			add_action( 'shutdown', [ $this, 'afterRegisterAllStringsInShutdown' ], ShutdownHooks::PRIORITY_REGISTER_STRINGS + 1 );
--- a/tests/phpunit/tests/AutoUpdate/TestHooks.php
+++ b/tests/phpunit/tests/AutoUpdate/TestHooks.php
@@ -18,6 +18,7 @@ class TestHooks extends  TestCase {
 
 		$subject = $this->getSubject();
 
+		\WP_Mock::expectFilterNotAdded( 'wpml_pb_auto_update_enabled', '__return_true' );
 		\WP_Mock::expectFilterNotAdded( 'wpml_tm_delegate_translation_statuses_update', [ $subject, 'enqueueTranslationStatusUpdate'], 10, 3 );
 		\WP_Mock::expectFilterNotAdded( 'wpml_tm_post_md5_content', [ $subject, 'getMd5ContentFromPackageStrings' ], 10, 2 );
 		\WP_Mock::expectActionNotAdded( 'shutdown', [ $subject, 'afterRegisterAllStringsInShutdown' ], \WPML\PB\Shutdown\Hooks::PRIORITY_REGISTER_STRINGS + 1 );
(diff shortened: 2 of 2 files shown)


andreasciamanna:
@strategio can you share a use case of this filter?

Without it, I would say that you just need to check for `defined( 'WPML_TM_VERSION' )` to do the same.

Also, provided that the filter is indeed needed, I would name it `wpml_pb_auto_update_active` (it has the same meaning, but it's shorter).

strategio:
> @strategio can you share a use case of this filter?
> 
> Without it, I would say that you just need to check for `defined( 'WPML_TM_VERSION' )` to do the same.

@andreasciamanna, I wanted a more specific check so that we can possibly enable or disable the feature in one place. When I had to temporarily disable the feature before WPML 4.4.0, I had to remind the change in the Elementor package.

So now, everything about the "translation auto-update" feature lies on the class `WPML\PB\AutoUpdate\Hooks`.

I was even thinking that we could offer an option to enable/disable and with this it would be very easy to implement.

andreasciamanna:
> @andreasciamanna, I wanted a more specific check so that we can possibly enable or disable the feature in one place. When I had to temporarily disable the feature before WPML 4.4.0, I had to remind the change in the Elementor package.

Ok, so the filter is meant to enable/disable the feature?
Where is this done?
Not here, as far as I can see.

Also, if the purpose is to enable/disable a feature, should one of these words be part of the filter name? 
Like `wpml_pb_auto_update_enabled` or `wpml_pb_auto_update_disabled` (depending on how you want to read and use it).

strategio:
@andreasciamanna, I changed the filter to `wpml_pb_auto_update_enabled` as you suggested. For me it looks like synonyms... I think we could even use `wpml_pb_auto_update_loaded` maybe.

When I say we can "atomically" disable it, I mean that we just need to comment out `WPML\PB\AutoUpdate::class` inside `\WPML_PB_Loader::load_hooks` (and we don't need to change code in any other place... the old translation status behavior will continue working as it used to do. Do it make sense to you?

andreasciamanna:
> When I say we can "atomically" disable it, I mean that we just need to comment out `WPML\PB\AutoUpdate::class` inside `\WPML_PB_Loader::load_hooks` (and we don't need to change code in any other place), the old translation status behavior will continue working as it used to do. Do it make sense to you?

I'm sorry, no.

That's why I asked to give me a use case, an example of how this would work.

This PR only has an `apply_filter`: I can't understand how this filter would be used.

strategio:
@andreasciamanna, here's the place where I used this filter https://github.com/OnTheGoSystems/wpml-page-builders-elementor/pull/123/files#diff-56085cc59e85b18171cc1c6a5a104794R127.

Right now, the only condition to "enable" the feature is that we have `WPML_TM_VERSION` defined. But could change in the future if we decide to have one option to change it in WPML settings for instance.

Here's the place where we load the "translation auto-update" feature (https://github.com/OnTheGoSystems/wpml-page-builders/blob/develop/src/st/class-wpml-pb-loader.php#L84). As you can see, it's just 1 line, everything of this feature lies in this class. If for some reason we need to "disable" the feature or add a new setting to do so, it's very easy and we won't have to make any change in other places (what I unfortunately had to do in Elementor when I had to revert the feature before WPML 4.4.0).

andreasciamanna:
So, the filter is useless at the moment, unless we add an option to disable the functionality.

Is that correct?

If so, I don't really see the point of adding it now.

strategio:
@andreasciamanna, you don't see the point of having modular (incremental) features in our code? Maybe I am pessimistic, but I can't tell we'll have 0 problem with this new feature when it will go live. If this ever happens, we can quickly give a patch (1 line to comment out), and later we can implement a new WPML settings (or some other solution). I hope we won't ever need to do this, but I believe this code organization make things easier to maintain.

Moreover, now we have one specific code in Elementor's package, but we might find another bug later with another component and we'll add custom code elsewhere. If we have this filter hook to check, it makes things easier.

To be complete, I initially tried to have the Elementor's code located (and generalized to all API page builders) in the `WPML\PB\AutoUpdate\Hooks` class but it was not possible without a massive refactoring, so I gave up.

Well, if you think my arguments are not valid, I will just revert the 2 reverted commits:
- https://github.com/OnTheGoSystems/wpml-page-builders/commit/a3f63b59e2b4a6c5caf6e7a1bb1532fbcd23d3f3
- https://github.com/OnTheGoSystems/wpml-page-builders-elementor/commit/440c70651c3ab54b20038a3365bf7545c683b531

andreasciamanna:
@strategio I'm in favour of incremental feature, but here I only see a filter which isn't used anywhere.

strategio:
>@strategio I'm in favour of incremental feature, but here I only see a filter which isn't used anywhere.

The filter callback is here in the main PB package, and the filter hook is in the Elementor PB package @andreasciamanna => https://github.com/OnTheGoSystems/wpml-page-builders-elementor/pull/123/files#diff-56085cc59e85b18171cc1c6a5a104794R127.

andreasciamanna:
@strategio 

> The filter callback is here in the main PB package, and the filter hook is in the Elementor PB package @andreasciamanna => https://github.com/OnTheGoSystems/wpml-page-builders-elementor/pull/123/files#diff-56085cc59e85b18171cc1c6a5a104794R127.

But is like not having it because it has no effect: `apply_filters( 'wpml_pb_auto_update_enabled', false )` always returns true.

There is not an incremental feature if that new feature has no effects.

I would accept the new filter if is was actually used to disable auto-update, but that's not the case.

strategio:
Here are the new PRs @andreasciamanna, please check it:
- https://github.com/OnTheGoSystems/wpml-page-builders/pull/112
- https://github.com/OnTheGoSystems/wpml-page-builders-elementor/pull/124

PR #111, opened 2020-09-10
Allow to reuse string translations in API Page Builders

strategio:
This is an optional feature that is only required for Cornerstone (see implementation in https://github.com/OnTheGoSystems/wpml-page-builders-cornerstone/pull/17).

The reason is that most API page builders are keeping a unique
identifier for each module, and this ID does not change if we make some
change to the module.

For Cornerstone, we don't have any ID and we use a hash built with the
serialized module instead. The consequence is that the ID changes every
time we make a change to the module (and we strings identifiers are not matching either).

Moreover, in `WPML_Page_Builders_Update_Translation`, we were applying the
translation only if it's completed and I removed it because:
1. We rely on the post translation status to apply the strings, not on
each string status.
2. We update the PB string translation only via a translation editor (not via the ST UI).

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-7551

### DIFF
--- a/src/st/compatibility/class-wpml-page-builders-update-translation.php
+++ b/src/st/compatibility/class-wpml-page-builders-update-translation.php
@@ -48,9 +48,7 @@ abstract class WPML_Page_Builders_Update_Translation extends WPML_Page_Builders_
 		if ( array_key_exists( $string->get_name(), $this->string_translations ) &&
 		     array_key_exists( $this->lang, $this->string_translations[ $string->get_name() ] ) ) {
 			$translation = $this->string_translations[ $string->get_name() ][ $this->lang ];
-			if ( (int) $translation['status'] === self::TRANSLATION_COMPLETE ) {
-				$string->set_value( $translation['value'] );
-			}
+			$string->set_value( $translation['value'] );
 		}
 
 		return $string;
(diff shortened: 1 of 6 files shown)


src/st/strategy/api-hooks/class-wpml-pb-api-hooks-strategy.php
```diff
@@ -58,7 +58,7 @@ public function get_content_updater() {
 	}
 
 	public function get_package_strings( $package_data ) {
-		$this->factory->get_string_translations( $this )->get_package_strings( $package_data );
+		return $this->factory->get_string_translations( $this )->get_package_strings( $package_data );
```
strategio, line 61: I believe we never really use this method in this strategy...

andreasciamanna (approved):

strategio:
@brucepearson, I'll let you check this one and also https://github.com/OnTheGoSystems/wpml-page-builders-cornerstone/pull/17 as you worked on this reuse feature.

src/st/compatibility/class-wpml-page-builders-update-translation.php
```diff
@@ -48,9 +48,7 @@ protected function get_translation( WPML_PB_String $string ) {
 		if ( array_key_exists( $string->get_name(), $this->string_translations ) &&
 		     array_key_exists( $this->lang, $this->string_translations[ $string->get_name() ] ) ) {
 			$translation = $this->string_translations[ $string->get_name() ][ $this->lang ];
-			if ( (int) $translation['status'] === self::TRANSLATION_COMPLETE ) {
```
brucepearson, line 51: @strategio Are you sure it's ok to remove this check

brucepearson: @strategio I just saw you comment so I guess it's ok.

brucepearson (approved):
@strategio Looks good. +1

PR #112, opened 2020-09-11
Introduce `WPML\PB\AutoUpdate\Settings` and constant `WPML_TRANSLATION_AUTO_UPDATE_ENABLED`

strategio:
This helps to keep an "atomic" logic for the whole feature and make it easy to disable if we
have any kind of trouble. We just need to define the constant in the
`wp-config.php` file.

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-7544

### DIFF
--- a/src/AutoUpdate/Settings.php
+++ b/src/AutoUpdate/Settings.php
@@ -1,0 +1,21 @@
+<?php
+
+namespace WPML\PB\AutoUpdate;
+
+class Settings {
+
+	/**
+	 * This is part of the "Translation Auto-Update" feature
+	 * The "Translation Auto-Update" feature will be released in the next major version.
+	 * We need a way for users to allow disabling it quickly, if necessary.
+	 *
+	 * @return bool
+	 */
+	public static function isEnabled() {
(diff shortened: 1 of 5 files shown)


src/AutoUpdate/Settings.php
```diff
@@ -0,0 +1,17 @@
+<?php
+
+namespace WPML\PB\AutoUpdate;
+
+class Settings {
```
andreasciamanna, line 5: @strategio remember to add a PHPDoc here with a description explaining the purpose of this class:

- This is part of the "AutoUpdate" feature
- The "AutoUpdate" feature will be released in the next major version. We need a way for users to allow disabling it quickly, if necessary.

strategio: Right, done!

andreasciamanna (approved):

strategio:
@OnTheGoSystems/compatibility, @OnTheGoSystems/wpml, one more review please.

brucepearson (approved):

PR #113, opened 2020-09-14
Introduce logic to extract and save API PB configuration from the general config file

strategio:
The `Hooks` claas will be implemented later in the API PB packages with a shared factory.

Please check the [reference for new allow items in the config file](https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-7560).

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-7561

### DIFF
--- a/src/Config/Parser.php
+++ b/src/Config/Parser.php
@@ -1,0 +1,158 @@
+<?php
+
+namespace WPML\PB\Config;
+
+use WPML\FP\Fns;
+use WPML\FP\Lst;
+use WPML\FP\Maybe;
+use WPML\FP\Obj;
+
+class Parser {
+
+	/** @var string $configRoot */
+	private $configRoot;
+
+	/** @var string $defaultConditionKey */
+	private $defaultConditionKey;
+
+	public function __construct( $configRoot, $defaultConditionKey ) {
+		$this->configRoot          = $configRoot;
+		$this->defaultConditionKey = $defaultConditionKey;
+	}
+
+	/**
+	 * Receives a raw config array (from XML) and convert it into
+	 * a page builder configuration array.
+	 *
+	 * @see WPML_Elementor_Translatable_Nodes::get_nodes_to_translate()
+	 *
+	 * @param array $allConfig
+	 *
+	 * @return array
+	 */
+	public function extract( array $allConfig ) {
+		$pbConfig   = [];
+		$allWidgets = Obj::pathOr( [], [ 'wpml-config', $this->configRoot, 'widget' ], $allConfig );
+
+		foreach ( $allWidgets as $widget ) {
+			$widgetName = Obj::path( ['attr', 'name'], $widget );
+
+			$pbConfig[ $widgetName ] = [
+				'conditions' => $this->parseConditions( $widget, $widgetName ),
+				'fields'     => $this->parseFields( $widget ),
+			];
+
+			$fieldsInItem = Obj::prop( 'fields-in-item', $widget );
+
+			if ( $fieldsInItem ) {
+				$itemOf                                    = Obj::path( [ 'attr', 'items_of' ], $fieldsInItem );
+				$pbConfig[ $widgetName ]['fields_in_item'] = [ $itemOf => $this->parseFieldsInItem( $fieldsInItem ) ];
+			}
+
+			$integrationClasses = $this->parseIntegrationClasses( $widget );
+
+			if ( $integrationClasses ) {
+				$pbConfig[ $widgetName ]['integration-class'] = $integrationClasses;
+			}
+		}
+
+		return $pbConfig;
+	}
+
+	/**
+	 * @param array  $widget
+	 * @param string $widgetName
+	 *
+	 * @return array
+	 */
+	private function parseConditions( array $widget, $widgetName ) {
+		$makePair = function( $condition ) {
+			return [ Obj::pathOr( $this->defaultConditionKey, ['attr', 'key'], $condition ), $condition['value'] ];
+		};
+
+		return Maybe::fromNullable( Obj::path( ['conditions', 'condition'], $widget ) )
+			->map( [ $this, 'normalize' ] )
+			->map( Fns::map( $makePair ) )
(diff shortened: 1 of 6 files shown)


src/Config/Parser.php
```diff
@@ -0,0 +1,172 @@
...
+	 *
+	 * @see WPML_Elementor_Translatable_Nodes::get_nodes_to_translate()
+	 *
+	 * @param array $allConfig
+	 *
+	 * @return array
+	 */
+	public function extract( array $allConfig ) {
+		$pbConfig = [];
+		$allWidgets = Obj::path( [ 'wpml-config', $this->configRoot, 'widget' ], $allConfig );
```
brucepearson, line 33: @strategio I think you can use `pathOr( [], .... )` here to return an empty array and then remove the `if( $allWidgets )` check

src/Config/Parser.php
```diff
@@ -0,0 +1,172 @@
...
+			$parsedConditions = [];
+
+			foreach ( $this->normalize( $rawConditions ) as $condition ) {
+				$parsedConditions[ Obj::pathOr( $this->defaultConditionKey, ['attr', 'key'], $condition ) ] = $condition['value'];
+			}
+		} else {
+			$parsedConditions = [ $this->defaultConditionKey => $widgetName ];
+		}
+
+		return $parsedConditions;
```
brucepearson, line 81: @strategio You could try something like this. You'll need to make `normalize` public for this.

```php
$makePair = function( $condition ) { 
    return [Obj::pathOr( $this->defaultConditionKey, ['attr', 'key'], $condition ), $condition['value'] ] 
};
return Maybe::fromNullable( Obj::path( ['conditions', 'condition'], $widget ) )
   ->map( [ $this, 'normalize ] )
   ->map( Fns::map( $makePair ) )
   ->map( Lst::fromPairs() )
   ->getOrElse(  [ $this->defaultConditionKey => $widgetName ] );
```

src/Config/Parser.php
```diff
@@ -0,0 +1,172 @@
...
+	}
+
+	/**
+	 * @param array $widget
+	 *
+	 * @return array
+	 */
+	private function parseFields( array $widget ) {
+		$rawFields    = Obj::path( ['fields', 'field'], $widget );
```
brucepearson, line 90: @strategio Maybe use pathOr with a default of an empty array.

src/Config/Parser.php
```diff
@@ -0,0 +1,172 @@
...
+		return $parsedFields;
+	}
+
+	/**
+	 * @param array $fieldsInItem
+	 *
+	 * @return array
+	 */
+	private function parseFieldsInItem( array $fieldsInItem ) {
+		$rawFieldsInItem    = Obj::prop( 'field', $fieldsInItem );
```
brucepearson, line 126: @strategio Maybe propOr here

src/Config/Parser.php
```diff
@@ -0,0 +1,172 @@
...
+	 */
+	private function parseIntegrationClasses( array $widget ) {
+		$rawClasses    = Obj::path( [ 'integration-classes', 'integration-class' ], $widget );
+		$parsedClasses = [];
+
+		if ( $rawClasses ) {
+			$parsedClasses = Lst::pluck( 'value', $this->normalize( $rawClasses ) );
+		}
+
+		return $parsedClasses;
```
brucepearson, line 158: @strategio 

```php
return Obj::pathOr( [] [ 'integration-classes', 'integration-class' ], $widget )
   ->map( [ $this, 'normalize' ] )
   ->map( Lst::pluck( 'value' );
```

brucepearson (changes requested):
@strategio The code looks good. I just made a few suggestions you might want to make.

tests/phpunit/tests/Config/TestStorage.php
```diff
@@ -0,0 +1,56 @@
...
+
+	const PB_KEY = 'the-pb-key';
+
+	/**
+	 * @test
+	 */
+	public function itShouldGet() {
+		$pbConfig = [ 'some PB config' ];
+
+		$optionManager = $this->getOptionManager();
```
brucepearson, line 21: @strategio You can use the `OptionMock` trait for testing and then you don't need to mock OptionManager. You can just  use the Store get and set function to test.

```
$subject = $this->getSubject( new OptionManager() );
$subject->update( $pbConfig );

$this->assertEquals( $pbConfig, $subject->get() );
```

strategio: @brucepearson, it's not so simple becaue [OptionManager](https://git.onthegosystems.com/wpml/sitepress-multilingual-cms/-/blob/develop/classes/options/class-wpml-option-manager.php) class is in Core.

src/Config/Parser.php
strategio: Done!

src/Config/Parser.php
strategio: Excellent @brucepearson , done!

src/Config/Parser.php
strategio: Done

src/Config/Parser.php
strategio: Done

src/Config/Parser.php
strategio: Adjusted and added.

strategio (commented):
Thanks @brucepearson! I was also planning to improve the code yesterday but I believe it was a bit late...

strategio:
Ooops, I forgot to push the last changes. This is ready now!

tests/phpunit/tests/Config/TestStorage.php
brucepearson: @strategio I forgot. I did have a plan to move this in WP library. I might do it sometime soon to make these things easier.

brucepearson (approved):
@strategio Looks good. +1

dgwatkins (approved):
:+1: @strategio

PR #116, opened 2020-09-17
Simplify/unify parsing of field in fields or fields-in-item

strategio:
The way we handle the `key_of` attribute in `fields` field is not
optimal because we use it a non-numeric key. But it's also used in
several places in our code (and derived PB packages) so I decided to
continue with the same pattern for `fields-in-item`. At least we will be
consistent and we'll have only one way to parse a field.

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-7565


dgwatkins (approved):
:+1: @strategio

PR #117, opened 2020-09-23
Don't escape already escaped chars

gordonsh24:
https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-5981

### DIFF
--- a/src/st/strategy/shortcode/class-wpml-pb-update-shortcodes-in-content.php
+++ b/src/st/strategy/shortcode/class-wpml-pb-update-shortcodes-in-content.php
@@ -195,7 +195,7 @@ class WPML_PB_Update_Shortcodes_In_Content {
 	 */
 	private function filter_attribute_translation( $translation, $encoding ) {
 		if ( 'allow_html_tags' !== $encoding ) {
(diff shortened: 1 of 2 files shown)


dgwatkins:
@gordonsh24 I feel we are just patching the problem this way. The origin of the problem must be further up.

I am not 100% sure about it so I really prefer @strategio to take a look in this case.

gordonsh24:
@dgwatkins the problem is that the attribute already contains escaped text like: `My text A &amp;  B and other`. When you transform it with `htmlspecialchars` `&` character will be escaped and we will get `My text A &amp;amp;  B and other`.

I don't see other ways of fixing this issue.

dgwatkins:
Does `My text A &amp; B ` come like this directly from the post_content?

If so, you have my approval.

gordonsh24:
@dgwatkins yes, it does. It is a text provided by the user as DIVI button caption.

dgwatkins:
:+1: @gordonsh24

dgwatkins (approved):

strategio (approved):
+1 @gordonsh24

PR #119, opened 2020-10-20
Allow multiple occurrences of `fields-in-item` in XML config

strategio:
This is a new data shape used in Ultimate Addons for Elementor and it was not supported yet.

multiple-fields-in-items

It will be now possible to support it with the XML config.

I also added a second commit to lower the priority of the XML config so it can overwrite the PHP config.

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-7337

### DIFF
--- a/src/Config/Hooks.php
+++ b/src/Config/Hooks.php
@@ -6,6 +6,8 @@ use function WPML\FP\tap as tap;
 
 class Hooks implements \IWPML_Action {
 
+	const PRIORITY_AFTER_DEFAULT = 20;
+
 	/** @var Parser $parser */
 	private $parser;
 
@@ -27,7 +29,7 @@ class Hooks implements \IWPML_Action {
 
 	public function add_hooks() {
 		add_filter( 'wpml_config_array', tap( [ $this, 'extractConfig' ] ) );
-		add_filter( $this->translatableWidgetsHook , [ $this, 'extendTranslatableWidgets' ] );
+		add_filter( $this->translatableWidgetsHook , [ $this, 'extendTranslatableWidgets' ], self::PRIORITY_AFTER_DEFAULT );
 	}
 
 	public function extractConfig( array $allConfig ) {
--- a/src/Config/Parser.php
+++ b/src/Config/Parser.php
@@ -42,11 +42,16 @@ class Parser {
 				'fields'     => $this->parseFields(  Obj::pathOr( [], ['fields', 'field'], $widget ) ),
 			];
 
-			$fieldsInItem = Obj::prop( 'fields-in-item', $widget );
+			$fieldsInItems = Obj::prop( 'fields-in-item', $widget );
 
-			if ( $fieldsInItem ) {
-				$itemOf                                    = Obj::path( [ 'attr', 'items_of' ], $fieldsInItem );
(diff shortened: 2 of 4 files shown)


brucepearson (dismissed):
@strategio Looks good. +1

src/Config/Hooks.php
```diff
@@ -6,6 +6,8 @@
 
 class Hooks implements \IWPML_Action {
 
+	const PRIORITY_AFTER_DEFAULT = 20;
```
andreasciamanna, line 9: @strategio I don't quite understand the meaning of this.

Do you mean a priority value that is after the default value of "10"?

Anyway, why you need to define it in a constant? As far as I can see, it's only used in one place (and in the test).

src/Config/Parser.php
```diff
@@ -130,6 +135,10 @@ private function parseIntegrationClasses( array $widget ) {
 	 * @return array
 	 */
 	public function normalize( array $partialConfig ) {
-		return isset( $partialConfig['value'] ) ? [ $partialConfig ] : $partialConfig;
+		$isAssocArray = function( array $array ) {
+			return count( array_filter( array_keys( $array ), 'is_string' ) ) > 0;
+		};
```
andreasciamanna, line 140: @strategio I don't see the point of overcomplicating the code here with a function.

Can't you just set the value of `$isAssocArray` to `count( array_filter( array_keys( $$partialConfig ), 'is_string' ) )`?


```php
public function normalize( array $partialConfig ) {
	$isAssocArray = count( array_filter( array_keys( $partialConfig ), 'is_string' ) )
	return $isAssocArray ? [ $partialConfig ] : $partialConfig;
}		
```

andreasciamanna (changes requested):

src/Config/Parser.php
strategio: Right @andreasciamanna, it's overkill... simplified as suggested.

andreasciamanna (approved):

src/Config/Hooks.php
strategio: Generally we don't like magic numbers. I could have chosen 11 or `PHP_INT_MAX`. What really matter in this priority is that it comes somewhere after 10 (default). IMHO, it helps to understand what we can do with this priority in the future.

andreasciamanna: My commend it about the name, Pierre.

But sometimes I think we are a bit too fundamentalist when it comes to what are simple guidelines 👍 

The other problem here is the name of the constant that I find confusing.

Priorities are "higher" or "lower".

As you pointed out, you could have used `PHP_INT_MAX` and remove the need to define a constant which is used only in one place.

Anyway, I've already approved this PR. I just wanted to reply your last comment :)

strategio: I think `PHP_INT_MAX` is not a proper usage here. Developers will use the default value (10) because they are just adding configuration. I don't know if there is a use case for them to use `PHP_INT_MAX` but we should keep this priority available if it's not a compulsory requirement for us.

We have this requirement sometimes, but we should have a good reason to use it IMO (see [a good example](https://github.com/Automattic/jetpack/issues/14757) where end up fighting with other devs :) ).

andreasciamanna: Good point!

PR #122, opened 2020-12-04
Transform class `WPML_PB_Last_Translation_Edit_Mode` to use only static methods

strategio:
**Goes with https://git.onthegosystems.com/wpml/sitepress-multilingual-cms/-/merge_requests/3021**

This class has no state and it was passed too early to
`WPML\Compatibility\WPBakery\Styles` as a dependency. Now it will just
be used as a static call there.

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-7852

Note: This class is already covered in `Test_WPML_PB_Last_Edit_Mode`.

### DIFF
--- a/src/st/class-wpml-pb-last-translation-edit-mode.php
+++ b/src/st/class-wpml-pb-last-translation-edit-mode.php
@@ -11,8 +11,8 @@ class WPML_PB_Last_Translation_Edit_Mode {
 	 *
 	 * @return bool
 	 */
-	public function is_native_editor( $post_id ) {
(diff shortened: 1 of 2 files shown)


dgwatkins (approved):
:+1:  @strategio

andreasciamanna:
@strategio I assume we will release wpml-page-builders as a plugin not before releasing 4.5.0, right?

Otherwise, this kind of change would cause quite some problems.

strategio:
Yes @andreasciamanna, that's for WPML 4.5.0.

PR #123, opened 2020-12-18
Introduce logic to cleanup GB meta data in post content when it's built with a shortcode page builder

strategio:
For now, this will be used only for Enfold Avia Builder ( [see MR3046 in Core](https://git.onthegosystems.com/wpml/sitepress-multilingual-cms/-/merge_requests/3046)), but we should be able to reuse it also for other shortcode page builders.

The code located in the new `WPML\PB\GutenbergCleanup\Utils` class is inspired from some code we already have in the Elementor package (see https://github.com/OnTheGoSystems/wpml-page-builders-elementor/pull/121/files#diff-039918ee390a9373a7ba3ff801eb19e629d0d7ac36df0eeeaa768372b8ed6aa1R37). I will create a separate ticket to refactor the code in Elementor and reuse the new pieces in this main package.

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-7870

### DIFF
--- a/src/GutenbergCleanup/ShortcodeHooks.php
+++ b/src/GutenbergCleanup/ShortcodeHooks.php
@@ -1,0 +1,63 @@
+<?php
+
+namespace WPML\PB\GutenbergCleanup;
+
+use WPML\FP\Fns;
+use WPML\FP\Maybe;
+use WPML\FP\Obj;
+use WPML\LIB\WP\Gutenberg;
+use function WPML\FP\pipe;
+use function WPML\FP\tap as tap;
+
+class ShortcodeHooks implements \IWPML_Backend_Action {
+
+	public function add_hooks() {
+		add_action(
+			'wp_insert_post',
+			Fns::withoutRecursion( Fns::noop(), [ $this, 'removeGutenbergFootprint' ] ),
+			10, 2
+		);
+	}
+
+	/**
+	 * @param int      $post_ID
+	 * @param \WP_Post $post
+	 */
+	public function removeGutenbergFootprint( $post_ID, \WP_Post $post ) {
+		// $isBuiltWithShortcodes :: \WP_Post -> bool
+		$isBuiltWithShortcodes = function( \WP_Post $post ) {
+			/**
+			 * @since WPML 4.4.9
+			 *
+			 * @param bool     false
+			 * @param \WP_Post $post
+			 */
+			return apply_filters( 'wpml_pb_is_post_built_with_shortcodes', false, $post );
+		};
+
+		// $hasGutenbergMetaData :: \WP_Post -> bool
+		$hasGutenbergMetaData = pipe(
+			Obj::prop( 'post_content' ),
+			Gutenberg::hasBlock()
+		);
+
+		// $removeHtmlComments :: \WP_Post -> \WP_Post
+		$removeHtmlComments = tap( pipe(
+			Obj::over( Obj::lensProp( 'post_content' ), Gutenberg::stripBlockData() ),
+			'wp_update_post'
+		) );
+
+		// $deleteGutenbergPackage :: \WP_Post -> void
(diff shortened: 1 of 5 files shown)


src/GutenbergCleanup/Utils.php
```diff
@@ -0,0 +1,47 @@
...
+		return (bool) preg_match( '/<!-- wp:/', $content );
+	}
+
+	/**
+	 * @param string $content
+	 *
+	 * @return string
+	 */
+	public static function stripHtmlComments( $content ) {
+		return preg_replace( '(<!--[^<]*-->)', '', $content );
```
kkarpieszuk, line 24: @strategio I've checked your regex. this will not remove comment like this:

```
<!-- wp:myblock block which will wrap all its content in <code> html tag -->
```

because of `<` inside of the comment block which is allowed, rare though

I tried and regex like `'/(<!--.*-->)/sUmg'` is working on multiline and handled `<` too.

kkarpieszuk: https://regex101.com/r/OEXpDN/1

strategio: @kkarpieszuk, I don't think `<!-- wp:myblock block which will wrap all its content in <code> html tag -->` is a real case or maybe I didn't understand what you mean.

Anyway, I have strengthen the test constraint with your scenario and tests are still passing as you can see:

https://github.com/OnTheGoSystems/wpml-page-builders/pull/123/commits/e5246c72084360675213feaa3ea3df923f6ab31c

strategio:
@OnTheGoSystems/compatibility, @OnTheGoSystems/wpml, please review!

src/GutenbergCleanup/ShortcodeHooks.php
```diff
@@ -0,0 +1,62 @@
...
+		);
+	}
+
+	/**
+	 * @param int      $post_ID
+	 * @param \WP_Post $post
+	 */
+	public function removeGutenbergFootprint( $post_ID, \WP_Post $post ) {
+		// $isBuiltWithShortcodes :: \WP_Post -> bool
+		$isBuiltWithShortcodes = function( \WP_Post $post ) {
```
brucepearson, line 27: @strategio Maybe:

```php
$isBuiltWithShortcodes = partial( 'apply_filters', 'wpml_pb_is_post_built_with_shortcodes', false );
```

src/GutenbergCleanup/Utils.php
```diff
@@ -0,0 +1,47 @@
...
+use WPML\FP\Relation;
+
+class Utils {
+
+	/**
+	 * @param string $content
+	 *
+	 * @return bool
+	 */
+	public static function hasGutenbergMetaData( $content ) {
```
brucepearson, line 14: @strategio I think this is the same as the helper in WPML\LIB\WP.

```
WPML\LIB\WP\Gutenberg::hasBlock( $content );
```

brucepearson (approved):
@strategio Looks good. +1

kkarpieszuk (dismissed):
+1

src/GutenbergCleanup/Utils.php
strategio: Thanks for reminding @brucepearson, I will reuse this and also move my `Utils::stripHtmlComments` to this library.

strategio: Done (see https://git.onthegosystems.com/wpml-packages/wp/-/merge_requests/36)

src/GutenbergCleanup/ShortcodeHooks.php
strategio: I also had this in mind but I decided to opt for the verbose form to make it more readable (and findable) for non-fp-literate people.

PR #127, opened 2021-03-18
Fix warnings due to `WPML_PB_Last_Translation_Edit_Mode` uses as an instance instead of the static methods

strategio:
I also forced the tests to fail on warning because I discovered the problem by chance, the pipeline was showing the green light.

@kkarpieszuk, this is about the class I shared this morning :)

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlpb-187

### DIFF
--- a/phpunit.xml
+++ b/phpunit.xml
@@ -5,6 +5,7 @@
 		convertErrorsToExceptions="true"
 		convertNoticesToExceptions="true"
 		convertWarningsToExceptions="true"
+		failOnWarning="true"
 >
 	<testsuites>
 		<testsuite name="wpml-pb-tests">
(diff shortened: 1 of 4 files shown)


dgwatkins:
Looks good @strategio 

I grepped a bit and found it phpstan stubs (in both TM and WCML):
https://git.onthegosystems.com/wpml/wpml-translation-management/-/blob/develop/tests/phpstan/stubs/page-builders.stub#L779

Maybe you want to take a look?

dgwatkins (approved):

strategio:
> I grepped a bit and found it phpstan stubs (in ST, TM and WCML):
> https://git.onthegosystems.com/wpml/wpml-translation-management/-/blob/develop/tests/phpstan/stubs/page-builders.stub#L779

@dgwatkins, I left the `get_last_translation_edit_mode` in the factory, just marked it as deprecated. We can still use the static or instance methods in real life. The problem is that now we can't properly mock if using the instance method.

kkarpieszuk (commented):
looks good

PR #129, opened 2021-04-16
Filter post content before copying post without translatable elements

kkarpieszuk:
- introducing filter wpml_pb_before_page_without_elements_post_content_copy in \WPML_PB_Handle_Post_Body::copy

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlpb-194

### DIFF
--- a/src/tm/class-wpml-pb-handle-post-body.php
+++ b/src/tm/class-wpml-pb-handle-post-body.php
@@ -31,7 +31,9 @@ class WPML_PB_Handle_Post_Body implements IWPML_Backend_Action, IWPML_Frontend_A
 	public function copy( $new_post_id, $original_post_id, array $fields ) {
 		$original_post = get_post( $original_post_id );
 		if ( $original_post && $this->page_builders_built->is_page_builder_page( $original_post ) && ! $this->job_has_packages( $fields ) ) {
-			wpml_update_escaped_post( [ 'ID' => $new_post_id, 'post_content' => $original_post->post_content ] );
+			$post_content =
+				apply_filters( 'wpml_pb_before_page_without_elements_post_content_copy', $original_post->post_content, $new_post_id, $original_post_id );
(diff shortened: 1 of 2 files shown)


dgwatkins:
@kkarpieszuk Why not use the existing action `wpml_pb_after_page_without_elements_post_content_copy`?

You have the post_id so you can fetch the content and update it.

kkarpieszuk:
@dgwatkins to avoid redundant sql UPDATE. Saving post in `wpml_update_escaped_post` and doing the same a line below is not the smartest way :)

dgwatkins (approved):
The code is good, I will approve.

But, as we discussed yesterday, I dont quite agree with the logic.
We are adding too much complexity to address the performance of an edge case.

kkarpieszuk:
adjusting value on display would me even less effective in terms of performance. Plus, we already have similar logic for custom fields (before we save them in TE we allow to change their values) so it is the same, but for post content. 

@strategio please provide another review

strategio:
@kkarpieszuk, I have the same feeling as @dgwatkins: A lot of complexity for something that has been reported by only one user yet.

First, I am not very keen on adding new hooks to this project. We already have a very convoluted code here, so adding more hooks will just worsen the situation.

Then, if you really want to alter the post data saved at this point, you can also use the native WP filter hook `wp_insert_post_data`, because the function `wpml_update_escaped_post` will call `wp_insert_post`.

I also think we should just convert the ID on the frontend (on-the-fly). This should be very simple and the performance hit is not significant IMO, we already use this approach in many other cases. And I believe not all pages will contain a reference to another post, so it's not affecting the whole site.

And looking at the ticket, I see that this change only addresses the case when a translation job has no strings. You will have to address the other case (with strings)... And I already mentioned you will face another glitch if the containing post is translated before the nested one...

All in all, I really think we should just convert the ID on the fly.

Sorry, I won't approve your PR.

PR #133, opened 2021-09-03
Improve the way we flag is a translation has been edited with a translation editor or not

strategio:
With the "translation auto-update" feature, we can have translations
automatically updated while editing the original.

- We need to compare the globablly edited post ID with the locally saved
post ID.
- We will also pass the translated post ID to the filter.

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-8858

### DIFF
--- a/src/st/class-wpml-pb-integration.php
+++ b/src/st/class-wpml-pb-integration.php
@@ -1,6 +1,7 @@
 <?php
 
 use \WPML\FP\Fns;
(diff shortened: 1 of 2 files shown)


strategio:

gordonsh24 (approved):
@strategio +1

brucepearson (approved):
@strategio +1

dgwatkins (approved):

